PR #1, opened 2016-10-28
add ConfigMap generate script

brancz:
This also contains the generated ConfigMaps for the grafana dashboards
as well as for alerts. These will be updated in subsequent changes as
sets of alerts and dashboards are added/updated in the respective
subdirectories in `/assets`.

In general this allows the dashboards to actually be reviewed as changes in the json are now actually visible. To be able to review them best it would be good if we always format them the same way. Right now I did this with `cat *-dashboard.json | python -m json.tool`, which would be good to document at some point so people who want to contribute know the procedure. Let me know if you think we should do that straight away.

@fabxc also let me know what you think about removing the etcd alerts for now. If you want I can add them back in, but then they will be included in the set of alerts in the ConfigMap, I figured that that might not be what we want right now.


fabxc:
Okay, we can remove etcd3 alerts for now. Theoretically, it shouldn't matter as the alerts will simply have no effect in absence of the used metrics.

Rest sounds good. Documentation is the most important part of this repository, I'd say. So we should have it.
We might want to split out from the main README into distinct ones eventually.

👍

brancz:
sounds like a plan 👍 don't have merge rights ... if you would be so kind 😁

fabxc:
You have now.

PR #2, opened 2016-11-01
monitor and alert on k8s components

brancz:
@fabxc


brancz:
etcd alerts coming up shortly after we get this in 🎉

brancz:
So the kube-scheduler and kube-controller-manager endpoints are already there by default but have some [expectations](https://github.com/kubernetes-incubator/bootkube/issues/64#issuecomment-228144345). I'll update this PR to create custom endpoints objects just to be safe.

PR #6, opened 2016-11-02
Extract prometheus.yaml into assets

grobie:
It's a lot easier to show and explain the Prometheus config if available
in the standard yaml format. The Kubernetes ConfigMap specs can always
be auto-generated.


assets/prometheus/prometheus.yaml
```diff
@@ -0,0 +1,70 @@
+global:
+  evaluation_interval: 30s
```
brancz: sgtm not sure why we did this initially

hack/scripts/generate-configmaps.sh
```diff
@@ -1,7 +1,10 @@
 #!/bin/bash
 
+# Generate Prometheus configuration ConfigMap
+kubectl create configmap --dry-run=true prometheus-k8s --from-file=assets/prometheus/ -oyaml > manifests/prometheus/prometheus-k8s-cm.yaml
```
brancz: Do you think there might be new files in this directory? If not, then just for the sake of understanding at first, sight we can select the config file directly in the `--from-file` flag instead of the entire directory. Feel free to ignore if you disagree.

brancz:
First outside contribution 👍 If you could just rebase then this lgtm

grobie:
@brancz rebased, made the path explicit as requested, unified scrape + evaluation interval.

brancz:
Thanks @grobie ! 👍

PR #7, opened 2016-11-03
Use node as job label for node-exporter metrics

grobie:
The Prometheus documentation and examples usually use job="node" for
node-exporter configurations, including console templates and several
Grafana dashboards.

@brancz @fabxc 

Other options:
- rename all examples from node to node-exporter. There are >4k results on Google for [job="node"](https://www.google.com/webhp#q=prometheus+job%3D%22node%22).
- rename the service to node

### DIFF
--- a/manifests/exporters/node-exporter-svc.yaml
+++ b/manifests/exporters/node-exporter-svc.yaml
@@ -5,6 +5,7 @@ metadata:
     app: node-exporter
   annotations:
     prometheus.io/scrape: 'true'
+    prometheus.io/job: node
   name: node-exporter
 spec:
   type: ClusterIP
--- a/manifests/prometheus/prometheus-k8s-cm.yaml
+++ b/manifests/prometheus/prometheus-k8s-cm.yaml
@@ -42,6 +42,10 @@ data:
         source_labels: [__meta_kubernetes_service_name]
         target_label: job
       - action: replace
+        source_labels: [__meta_kubernetes_service_annotation_prometheus_io_job]
+        target_label: job
+        regex: (.+)
+      - action: replace
         source_labels: [__meta_kubernetes_service_name]
         regex: kubernetes
         target_label: __scheme__
--- a/manifests/prometheus/prometheus-k8s-rules.yaml
+++ b/manifests/prometheus/prometheus-k8s-rules.yaml
@@ -1,57 +1,127 @@
 apiVersion: v1
 data:
-  etcd2.rules: "### General cluster availability ###\n\n# alert if another failed
-    peer will result in an unavailable cluster\nALERT InsufficientPeers\n  IF count(up{job=\"etcd-k8s\"}
-    == 0) > (count(up{job=\"etcd-k8s\"}) / 2 - 1)\n  FOR 3m\n  LABELS {\n    severity
-    = \"critical\"\n  }\n  ANNOTATIONS {\n    summary = \"Etcd cluster small\",\n
-    \   description = \"If one more etcd peer goes down the cluster will be unavailable\",\n
-    \ }\n\n### HTTP requests alerts ###\n\n# alert if more than 1% of requests to
-    an HTTP endpoint have failed with a non 4xx response\nALERT HighNumberOfFailedHTTPRequests\n
-    \ IF sum by(method) (rate(etcd_http_failed_total{job=\"etcd-k8s\", code!~\"4[0-9]{2}\"}[5m]))\n
-    \   / sum by(method) (rate(etcd_http_received_total{job=\"etcd-k8s\"}[5m])) >
-    0.01\n  FOR 10m\n  LABELS {\n    severity = \"warning\"\n  }\n  ANNOTATIONS {\n
-    \   summary = \"a high number of HTTP requests are failing\",\n    description
-    = \"{{ $value }}% of requests for {{ $labels.method }} failed on etcd instance
-    {{ $labels.instance }}\",\n  }\n\n# alert if more than 5% of requests to an HTTP
-    endpoint have failed with a non 4xx response\nALERT HighNumberOfFailedHTTPRequests\n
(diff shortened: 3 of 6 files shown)


manifests/exporters/node-exporter-svc.yaml
```diff
@@ -5,6 +5,7 @@ metadata:
     app: node-exporter
   annotations:
     prometheus.io/scrape: 'true'
+    prometheus/job_name: node
```
fabxc: `.io` missing.

How about just `prometheus.io/job`, `job_name` is really only specific to the Prometheus scrape config whereas the result is the plain `job` label.

fabxc:
I think the annotation is the right way to go. It should eventually be a first-class concept in the operator though. 

Currently the operator always appends `-<port>` though. We might want to drop it if there's just a single one.

manifests/exporters/node-exporter-svc.yaml
grobie: If you envision the operator itself supporting such annotations in the future, we could go for `monitoring.coreos.io/prometheus_job` or so.

manifests/prometheus/prometheus-k8s-cm.yaml
```diff
@@ -42,6 +42,10 @@ data:
         source_labels: [__meta_kubernetes_service_name]
         target_label: job
       - action: replace
+        source_labels: [__meta_kubernetes_service_annotation_prometheus_io_job]
+        target_label: job
+        regex: (.+)
```
grobie: Unfortunately the Prometheus default for regex is `(.*)`, which means if the annotation doesn't exist it would set job="".

fabxc:
@brian-brazil, I'd like to pull you in on the discussion of the annotation. (In the static configuration for cluster components here it's generally fine, ofc.)

Service names are unique (within a namespace) but several services can totally refer to the same application. The `job` label in Prometheus should be the same in this case.
In Kubernetes the semi-standardized `app` label is equivalent but we cannot rely on it in any way. 

I want to avoid dragging a lot of information out of the TPRs back into service annotations over time. Another way would be to specify in a `ServiceMonitor` a service label that should be used as `job` label.

Example:

```yaml
apiVersion: monitoring.coreos.com/v1alpha1
kind: ServiceMonitor
metadata:
  name: frontend
  labels:
    tier: frontend
spec:
  # ...
  # Service label of which we take the value and write it into `job`.
  jobLabel: app 
```

@grobie @matthiasr as proponents of pod-based monitoring: worth thinking about how we would find a job label when monitoring pods, i.e. introduce a `PodMonitor` TPR. We'd probably define a pod label being used as `job` label then as we don't have an external grouping defining it?
In that case defining a service label to `job` mapping in `ServiceMonitor` would seem more consistent.

@brancz

fabxc:
This requires fixing of etcd allerts in assets/rules to match any changed job names and running the asset generation script.

grobie:
@fabxc updated the rules/cm. Not sure why my kubectl generates the configmap differently (GitVersion:"v1.4.1").

brian-brazil:
That sounds to me like the default should be to use the annotation, then `app` if present, and finally fall back to the `service`.

manifests/prometheus/prometheus-k8s-rules.yaml
```diff
@@ -1,57 +1,127 @@
...
+        severity = "warning"
+      }
+      ANNOTATIONS {
+        summary = "slow HTTP requests",
+        description = "on ectd instance {{ $labels.instance }} HTTP requests to {{ $label.method }} are slow",
+      }
+
+    ### File descriptor alerts ###
+
+    instance:fd_utilization = process_open_fds / process_max_fds
```
brian-brazil: instance:process_open_fds_per_process_max_fds:ratio is the convention.

manifests/prometheus/prometheus-k8s-rules.yaml
```diff
@@ -1,57 +1,127 @@
...
+    # alert if file descriptors are likely to exhaust within the next 4 hours
+    ALERT FdExhaustionClose
+      IF predict_linear(instance:fd_utilization[1h], 3600 * 4) > 1
+      FOR 10m
+      LABELS {
+        severity = "warning"
+      }
+      ANNOTATIONS {
+        summary = "file descriptors soon exhausted",
+        description = "{{ $labels.job }} instance {{ $labels.instance }} will exhaust in file descriptors soon",
```
brian-brazil: s/in/its/

brancz:
@grobie interesting that your kubectl generates it this way. I generated these manifests using `v1.4.5`.

brancz:
I wouldn't use the `app` label at all it's too loose of a standard and as a user it would confuse me more if it is transparently used, while a prometheus namespaced annotation makes me realize the purpose and it works on services just as well as on pods. I think I prefer the annotation over a field in the TPR.

fabxc:
Mh, what bothers me a bit is that now I've to add some (repetitive) monitoring knowledge into all my services again – just like before with the other `prometheus.io/` annotations. When the original idea was in fact, that monitoring just works without even thinking about it if I stick to known standards for a group of services (typically enforced by a used language framework plus the CD pipeline).

In reality, if I've a group of services that share a common pattern of <port,path> combinations to expose metrics they can generally expected to follow the same label conventions – labels are how I group these services together to begin with. The example above reflects that.
Just reading `app` by default if the annotation is not there is indeed to loose. It would have to be specified in the TPR in some way.

brancz:
Fair enough, but how do we handle it here where we are not working with the ServiceMonitors/PodMonitors?

fabxc:
As we are writing our own config, anything is fair game.
Even if we go with the label mapping in the TPR for the generated configs of the operator, we can still think about respecting the annotation as a higher priority overwrite there.

brancz:
This was addressed partially by the prometheus-operator generating it's config and using the `jobLabel` field in the TPR respective to the `Service`'s labels.

PR #8, opened 2016-11-04
Scrape Prometheus service itself

grobie:
@fabxc or am I missing something you have in your setup?


fabxc:
I kept it out initially as I wanted to build out a dedicated meta-monitoring story.
However, with the governing `prometheus` service we originally only use for PetSets, this should unintentionally be solved by default.

I think keeping it within the `prometheus-k8s` is a sane choice for anyone using the default configuration.

PR #11, opened 2016-11-04
Fix always firing alerts

fabxc:
@brancz 

All aggregations using `by(cluster)` don't apply to our default setup. They come from a scenario were multiple clusters were monitored and custom target labels were attached for each.

Left them on there for now as they don't hurt. We might be able to use them if we ever build out a cross-cluster monitoring story.


brancz:
LGTM 👍

PR #12, opened 2016-11-04
Automatically deploy kube-system services

grobie:


brancz:
Not sure we should always do this. It expects the kubernetes components to be started in pods and having the correct labels. I mean they can't hurt, but might be confusing when there are not actually any matching pods.

fabxc:
Yea, the headless services we create are supposed to be the common denominator for things that might actually vary across clusters (without forcing the user to modify this in harder to understand areas like the Prometheus configuration). We cannot really do the right thing here beyond providing examples for cases like bootkube.

grobie:
You could test for the availability of in-cluster kube-controller-manager options.

```bash
if [[ "$(kctl --namespace=kube-system get pods --no-headers -lk8s-app=kube-controller-manager | wc -l)" -gt 0 ]]; then
  kctl -n=kube-system apply -f manifests/k8s
fi
```

I'll close this though. While it would make for a more seamless experience, a one-time one-liner is not that bad in the end.

PR #13, opened 2016-11-04
Fix broken kubernetes.rules

grobie:
The last change was lacking a closing parenthesis @brancz


brancz:
Thanks! 👍 I'm wondering if we can come up with a sane way we could do a simple end-to-end test to see if a change would introduce a config that for example could not be loaded. Maybe it would be enough to use promtool for this, maybe something more sophisticated.

PR #16, opened 2016-12-02
Update manifests to latest version of Prometheus Operator

brancz:
Tried out with a container image of `HEAD` of the Prometheus Operator.

List of changes divided into commits:
* Set lower resource requests by default (simply for demo purposes, as this currently won't work on minikube for example with only 1Gb of memory)
* Remove unnecessary prometheus.io/scrape annotation: (tl;dr) more confusing then helpful
* Add Alertmanager support in TPRs and manifests themselves (recently added to Prometheus Operator)
* Add kube-dns service for skydns and dnsmasq metrics to be collected. This change reflects work [upstream](https://github.com/kubernetes/kubernetes/blob/545f749a0deb8535656ce93ac1f35a59887839bf/cluster/addons/dns/skydns-rc.yaml.sed) and on [bootkube](https://github.com/kubernetes-incubator/bootkube/pull/179).
* Update `node-exporter` version

It would probably make sense to cut another release of the Prometheus Operator, just so we can at least partially version this until we get versioned user-guides.

@fabxc

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,10 +43,6 @@ To tear it all down again, run:
 hack/cluster-monitoring/teardown
 ```
 
-> All services in the manifest still contain the `prometheus.io/scrape = true`
-> annotations. It is not used by the Prometheus Operator. They remain for
-> pre Prometheus v1.3.0 deployments as in [this example configuration](https://github.com/prometheus/prometheus/blob/6703404cb431f57ca4c5097bc2762438d3c1968e/documentation/examples/prometheus-kubernetes.yml).
-
 ## Monitoring custom services
 
 The example manifests in [/manifests/examples/example-app](/manifests/examples/example-app)
--- a/assets/prometheus/prometheus.yaml
+++ b/assets/prometheus/prometheus.yaml
@@ -65,4 +65,4 @@ scrape_configs:
     regex: "kube-(.*)-prometheus-discovery"
   - action: keep
     source_labels: [__meta_kubernetes_endpoint_port_name]
-    regex: "prometheus"
+    regex: "prometheus.*"
--- a/manifests/alertmanager/alertmanager-config.yaml
+++ b/manifests/alertmanager/alertmanager-config.yaml
@@ -1,0 +1,18 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: alertmanager-main
+data:
+  alertmanager.yaml: |-
+    global:
+      resolve_timeout: 5m
+    route:
+      group_by: ['job']
+      group_wait: 30s
+      group_interval: 5m
+      repeat_interval: 12h
+      receiver: 'webhook'
+    receivers:
+    - name: 'webhook'
+      webhook_configs:
+      - url: 'http://alertmanagerwh:30500/'
--- a/manifests/alertmanager/alertmanager-service.yaml
+++ b/manifests/alertmanager/alertmanager-service.yaml
@@ -1,0 +1,14 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: alertmanager-main
+spec:
+  type: NodePort
+  ports:
+  - name: web
+    nodePort: 30903
+    port: 9093
+    protocol: TCP
+    targetPort: web
+  selector:
+    alertmanager: alertmanager-main
--- a/manifests/alertmanager/alertmanager.yaml
+++ b/manifests/alertmanager/alertmanager.yaml
@@ -1,0 +1,9 @@
+apiVersion: "monitoring.coreos.com/v1alpha1"
+kind: "Alertmanager"
+metadata:
+  name: "alertmanager-main"
+  labels:
+    alertmanager: "main"
+spec:
+  replicas: 3
+  version: v0.5.1
--- a/manifests/etcd/etcd-bootkube-gce.yaml
+++ b/manifests/etcd/etcd-bootkube-gce.yaml
@@ -5,8 +5,6 @@ metadata:
   labels:
     app: etcd
     etcd: k8s
-  annotations:
-    prometheus.io/scrape: 'true'
 spec:
   type: ClusterIP
   clusterIP: None
--- a/manifests/etcd/etcd-bootkube-vagrant-multi.yaml
+++ b/manifests/etcd/etcd-bootkube-vagrant-multi.yaml
@@ -5,8 +5,6 @@ metadata:
   labels:
     app: etcd
     etcd: k8s
-  annotations:
-    prometheus.io/scrape: 'true'
 spec:
   type: ClusterIP
   clusterIP: None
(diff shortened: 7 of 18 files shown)


fabxc:
👍 looks good.

Yes, we should cut another version to integrate into Tectonic etc.

julianvmodesto:
@brancz sorry if this isn't the best place to ask!

Regarding 0c1bb5c, why were prometheus.io/scrape annotations removed? Is there any way to configure Prometheus Operator to discover pods to scrape based on these annotations still?

fabxc:
The annotation is not supported by the operator. ServiceMonitors should be
used instead. The annotation is very limited by its nature and there's no
feasible way to support anything beyond "scrape on" and a single port. A
per-service scrape interval, multi-port pods, authentication, honorLabels,
and more are all not possible. It also provides no meaningful way to assign
services to be scraped by different Prometheus instances.

ServiceMonitors allow all these without exposing any complex relabelling
and there's really no practical reason for sticking with the annotation.

julianvmodesto:
Got it – this is helpful and makes sense after consuming, thank you!

cornelius-keller:
The problem I see here is that this annotations are fully self service and a well working way to communicate the metrics that should be scraped to prometheus. With ServiceMonitors this self service is not working anymore as they need to be in the monitoring namespace and I don't want to give everybody access to it.
As discussed here: https://github.com/coreos/prometheus-operator/pull/813 .
So for me regarding ease of setup etc. prometheus-operator is a great step, regarding the self service kapabilities that my previeous setup had it is a step back.

brancz:
The ServiceMonitor has a namespaceselector for the services, so the only thing you to do is have a convention in your organization, which boils down to the same as the annotation. You simply need you organization to have a commonly named port and a label like "monitoring: true". The namespace selector can then select Services globally.

brancz:
However, we acknowledge the want that people have, and are trying to think of a way we can introduce something more broad without violating the concerns we currently have.

StevenACoffman:
We have an established convention, but also often have third party tools installed which have their own conflicting conventions. As a hybrid solution, we are applying [a custom prometheus](https://github.com/Yolean/kubernetes-monitoring#custom-prometheus) on top of the operator. Thanks again @solsson

maver1ck:
I'd like to add one more thing to this issue.
There is a lot of 3rd party tools and ready to use helm charts that are using annotations.
And we can't use it out of the box.

StevenACoffman:
That is exactly our issue. 3rd party helm apps often have this need, which is in conflict with our using the operator at present.

maver1ck:
I found following solution.
In prometheus helm chart configuration add this:
```
  additionalScrapeConfigs:
  - job_name: 'kubernetes-pods'
    kubernetes_sd_configs:
    - role: pod
    relabel_configs:
    - source_labels: [__meta_kubernetes_pod_annotation_prometheus_io_scrape]
      action: keep
      regex: true
    - source_labels: [__meta_kubernetes_pod_annotation_prometheus_io_path]
      action: replace
      target_label: __metrics_path__
      regex: (.+)
    - source_labels: [__address__, __meta_kubernetes_pod_annotation_prometheus_io_port]
      action: replace
      regex: ([^:]+)(?::\d+)?;(\d+)
      replacement: $1:$2
      target_label: __address__
    - action: labelmap
      regex: __meta_kubernetes_pod_label_(.+)
    - source_labels: [__meta_kubernetes_namespace]
      action: replace
      target_label: kubernetes_namespace
    - source_labels: [__meta_kubernetes_pod_name]
      action: replace
      target_label: kubernetes_pod_name
```

suppix:
@maver1ck Could you please tell me where I should put this configuration ?

maver1ck:
You need to add this to helm chart configuration (values.yaml)
https://github.com/coreos/prometheus-operator/blob/master/helm/prometheus/values.yaml#L353

absolutejam:
> I found following solution.
> In prometheus helm chart configuration add this:
> 
> ```
>   additionalScrapeConfigs:
>   - job_name: 'kubernetes-pods'
>     kubernetes_sd_configs:
>     - role: pod
>     relabel_configs:
>     - source_labels: [__meta_kubernetes_pod_annotation_prometheus_io_scrape]
>       action: keep
>       regex: true
>     - source_labels: [__meta_kubernetes_pod_annotation_prometheus_io_path]
>       action: replace
>       target_label: __metrics_path__
>       regex: (.+)
>     - source_labels: [__address__, __meta_kubernetes_pod_annotation_prometheus_io_port]
>       action: replace
>       regex: ([^:]+)(?::\d+)?;(\d+)
>       replacement: $1:$2
>       target_label: __address__
>     - action: labelmap
>       regex: __meta_kubernetes_pod_label_(.+)
>     - source_labels: [__meta_kubernetes_namespace]
>       action: replace
>       target_label: kubernetes_namespace
>     - source_labels: [__meta_kubernetes_pod_name]
>       action: replace
>       target_label: kubernetes_pod_name
> ```
> Then all pods with prometheus.io/scrape annotations are processed.
> You don't need custom prometheus for that :)

There is also some information available at [https://github.com/coreos/prometheus-operator/blob/master/Documentation/additional-scrape-config.md](https://github.com/coreos/prometheus-operator/blob/master/Documentation/additional-scrape-config.md)

jmvizcainoio:
you can also add this config to scrape services

```
      - job_name: 'kubernetes-service'
        kubernetes_sd_configs:
        - role: service
        relabel_configs:
        - source_labels: [__meta_kubernetes_service_annotation_prometheus_io_scrape]
          action: keep
          regex: true
        - source_labels: [__meta_kubernetes_service_annotation_prometheus_io_path]
          action: replace
          target_label: __metrics_path__
          regex: (.+)
        - source_labels: [__address__, __meta_kubernetes_service_annotation_prometheus_io_port]
          action: replace
          regex: ([^:]+)(?::\d+)?;(\d+)
          replacement: $1:$2
          target_label: __address__
        - action: labelmap
          regex: __meta_kubernetes_service_label_(.+)
        - source_labels: [__meta_kubernetes_namespace]
          action: replace
          target_label: kubernetes_namespace
        - source_labels: [__meta_kubernetes_service_name]
          action: replace
          target_label: kubernetes_service_name
```

JulienBreux:
Update about `values.yaml` location; https://github.com/helm/charts/blob/master/stable/prometheus-operator/values.yaml

bchanson:
If you're compiling the yamls using the jsonnet file. Follow the [instructions mentioned earlier](https://github.com/coreos/prometheus-operator/blob/master/Documentation/additional-scrape-config.md) to create the additional scrape config secret. 

I tried to patch the prometheus resource directly after building/applying the yamls via jsonnet. It didn't work. I had to add the `additionalScrapeConfigs` config to the jsonnet file. 

This does not go in the `_config+::` section. It goes in `prometheus+::` like the serviceMonitor definitions.

```jsonnet
    prometheus+:: {
      prometheus+: {
        spec+: {
          additionalScrapeConfigs+: {
            name: 'additional-scrape-configs',
            key: 'additonal-scrape-configs.yaml'
          },
        },
      },
    },
```

PR #17, opened 2016-12-12
Fixed docker image tag for "quay.io/prometheus/node-exporter". 

bernielomax:
The docker image tag for the node exporter ("quay.io/prometheus/node-exporter") was not valid so I changed it to "v0.13.0".


brancz:
Thanks 👍

PR #19, opened 2016-12-13
correctly discover the Alertmanager cluster and ServiceMonitors

brancz:
Not sure how I missed these earlier but worth trying out I would say.

@fabxc @alexsomesan

### DIFF
--- a/manifests/examples/example-app/prometheus-frontend.yaml
+++ b/manifests/examples/example-app/prometheus-frontend.yaml
@@ -7,10 +7,9 @@ metadata:
     prometheus: frontend
 spec:
   version: v1.4.1
-  serviceMonitors:
-  - selector:
-      matchLabels:
-        tier: frontend
+  serviceMonitorSelector:
+    matchLabels:
+      tier: frontend
   resources:
     requests:
       # 2Gi is default, but won't schedule if you don't have a node with >2Gi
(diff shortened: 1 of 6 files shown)


manifests/examples/example-app/prometheus-frontend.yaml
```diff
@@ -7,10 +7,9 @@ metadata:
     prometheus: frontend
 spec:
   version: v1.4.1
-  serviceMonitors:
-  - selector:
-      matchLabels:
-        tier: frontend
+  serviceMonitorSelector:
+    matchExpressions:
+    - {key: tier, operator: In, values: [frontend]}
```
fabxc: Any reason for changing to expression matching. In the general case plain label matching seems simpler.

fabxc:
👍 

At some point we have to investigate what needs to be added to allow the k8s Prometheus to also be purely generated.

manifests/examples/example-app/prometheus-frontend.yaml
brancz: fixed

brancz:
Yes I was thinking of exactly the same thing. I will investigate tomorrow what we can get out of self-hosted today and what we still need.

brancz:
@fabxc as a follow-up there will be a point where all components will be self-hosted, so we can already come up with changes that will replace this manually maintained prometheus config with a replacement using the `ServiceMonitor`. This will come relatively soon, what's left is finishing work on bootkube regarding self hosted etcd, flannel and improvements on checkpointing.

PR #20, opened 2016-12-13
add ClusterRoleBinding as seen in CoreOS Tectonic

brancz:
@fabxc @alexsomesan


brancz:
This will be documented and handled by the prometheus-operator itself. Closing here.

PR #25, opened 2016-12-23
*: cleanup and unify minikube and self-hosted discovery

brancz:
This puts us very close to generating the config with `ServiceMonitor`s. Besides a lot of unification and cleanup this upgrades to `v0.2.1` of the prometheus operator so from now on `v1.5.x` clusters are required for the `hack/` scripts. Not that the content of this repo is now specific to `v1.5.x` it is just targeted by the operator, for `v1.4.x` clusters simply use the `v0.1.1` of the operator all other content should work the same way, but untested.

@fabxc @alexsomesan

Happy Holidays 🎉 🎅 🎄 🎁


fabxc:
👍

PR #26, opened 2016-12-30
Don't pass $KUBECONFIG to kubectl --kubeconfig flag

chancez:
The $KUBECONFIG variable is like $PATH, and supports multiple files
separated by `:`, but --kubeconfig only takes a single file as a value.
Since kubectl picks up the $KUBECONFIG variable already, don't pass it
to kubectl.

Also, use --namespace instead of -n to support older kubectl versions.


brancz:
thanks!

Issue #24, opened 2016-12-22
No kind "Prometheus" is registered


genti-t:
I'm trying to deploy prometheus using the manifests on this repo.

the TPRs are created
```
alertmanager.monitoring.coreos.com      Managed Alertmanager cluster          v1alpha1
prometheus.monitoring.coreos.com        Managed Prometheus server             v1alpha1
service-monitor.monitoring.coreos.com   Prometheus monitoring for a service   v1alpha1
```

but when I try to apply manifests/prometheus/prometheus-k8s.yaml :
```
 no kind "Prometheus" is registered for version "monitoring.coreos.com/v1alpha1"
```

the same goes for:
```
error: unable to decode "manifests/alertmanager/alertmanager.yaml": no kind "Alertmanager" is registered for version "
```

Can someone address me how to figure it  out why this is happening ?

brancz:
This error is relatively arbitrary, if you change the the `kubectl apply ...` to `kubectl create ...` it works. I will update the `hack/` scripts as a follow-up of this. Thanks for reporting!

brancz:
This only happens with a client and server version of `>=1.5.0`. Be aware that there is currently a problem with `emptyDir` volumes as described [here](https://github.com/coreos/prometheus-operator/issues/74).

Out of curiosity, do you have a working Persistent Volume provisioner for your cluster?

genti-t:
Thank you @brancz 
I have a new fresh cluster v.1.5.1. I have not tried any use for Persistent Volume on this version yet.

brancz:
In which environment if I may ask? GCE, AWS, base-metal? Is there a chance that you will have a PV Provisioner?

genti-t:
bare-metal and vagrant for now, and yes I am planing to use Persistent Volume provisioner

brancz:
We would greatly appreciate your help if you have it set up. Collecting reports in https://github.com/coreos/prometheus-operator/pull/75. Thanks and happy holidays 🙂

brancz:
Was the initial problem solved by not using `apply` @genti-t ? If so I think we can close this issue.

chancez:
I also saw this when running `kubectl apply`. It might be worth adding some comments into `hack/cluster-monitoring` scripts indicating why you use `kubectl create` and not `kubectl apply`. 

[...]

PR #27, opened 2017-01-03
hack/*: use kubectl apply where possible

brancz:
There is an issue with kubectl apply and third party resources, so for
the time being we use create for TPRs, but apply for all manifests that
it can handle.

Based on discussion in #24.

@chancez @genti-t @fabxc @alexsomesan

### DIFF
--- a/hack/example-service-monitoring/deploy
+++ b/hack/example-service-monitoring/deploy
@@ -4,4 +4,8 @@ if [ -z "${KUBECONFIG}" ]; then
     KUBECONFIG=~/.kube/config
 fi
 
-kubectl --kubeconfig="$KUBECONFIG" create -f manifests/examples/example-app
+kubectl --kubeconfig="$KUBECONFIG" apply -f manifests/examples/example-app/prometheus-frontend-svc.yaml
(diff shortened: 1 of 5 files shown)


hack/example-service-monitoring/deploy
```diff
@@ -4,4 +4,8 @@ if [ -z "${KUBECONFIG}" ]; then
     KUBECONFIG=~/.kube/config
 fi
 
-kubectl --kubeconfig="$KUBECONFIG" create -f manifests/examples/example-app
\ No newline at end of file
+kubectl --kubeconfig="$KUBECONFIG" apply -f manifests/examples/prometheus-frontend-svc.yaml
+kubectl --kubeconfig="$KUBECONFIG" apply -f manifests/examples/prometheus-frontend.yaml
```
alexsomesan: There is a Prometheus TPR in this file. Would `apply` actually work here?

brancz: you're absolutely right, my bad

fabxc:
👍

alexsomesan:
👍

chancez:
👍

PR #28, opened 2017-01-03
hack/cluster-monitoring: hide expected errors and print meaning

brancz:
When trying out `kube-prometheus` and running the `hack/` scripts and see errors, it feels like the user is doing something incorrectly, while those are expected errors when deploying the operator for the first time.

@fabxc @alexsomesan


fabxc:
👍

Issue #39, opened 2017-02-16
Grafana 4.1.1


tsloughter:
Is there a reason the version of grafana used is 3.1.1 still? I'm going to try with 4.1.1, as it appears to be needed for some dashboards I'm trying to use, but wanted to check if there are known issues with upgrading from 3.1.1.

brancz:
We have not attempted an update so far, simply because we didn't have the need for any new features, so if the upgrade is successful we're happy to merge it.

urzds:
I think this can be closed? #40 was already merged.

PR #40, opened 2017-02-16
upgrade grafana image to 4.1.1

tsloughter:
For #39 I tried it out and haven't had a problem.


brancz:
Thanks! I'm currently traveling so may not get to properly test this until monday, but your effort is much appreciated.

brancz:
The new version actually looks a bit more slick. I validated that all our templating and current dashboards are working, so lgtm :+1:

PR #46, opened 2017-02-21
fix memory/disk gauges and use rate

### LINKED ISSUE #45: Node's "free memory" in Grafana actually shows used memory
I installed Prometheus using the `deploy` script from 3a38e173f9a09f27d3a5a59c0cbd59ff8c671ba8.

When I access Grafana at http://MY_NODE:30902/dashboard/db/nodes?var-server=MY_NODE:9100, it displays "Free memory". However, the numbers seemed way off (displaying 5GiB free on a machine that runs almost no services and should have 24GiB of RAM), so I compared with the output of `free -m` on that node. Grafana appears to actually show the "used memory", not "free memory".

Additionally: Maybe adding another graph to the plot, showing "total memory", would help putting the numbers into context.

brancz:
As reported in #45 the some labels were off as they said "free memory" although memory usage was shown on the chart, and the gauge was actually showing the free memory, which is confusing for a gauge. Same applied to disk charts. Additionally prefer `rate` over `irate` as they should only really spike after ~3 scrapes which is ~2 minutes.

Fixes #45 

@fabxc 

/cc @urzds

### DIFF
--- a/assets/grafana/node-dashboard.json
+++ b/assets/grafana/node-dashboard.json
@@ -100,7 +100,7 @@
           "steppedLine": false,
           "targets": [
             {
-              "expr": "100 - (avg by (cpu) (irate(node_cpu{mode=\"idle\", instance=~\"$server\"}[5m])) * 100)",
+              "expr": "100 - (avg by (cpu) (irate(node_cpu{mode=\"idle\", instance=\"$server\"}[5m])) * 100)",
               "hide": false,
               "intervalFactor": 10,
               "legendFormat": "{{cpu}}",
@@ -188,7 +188,7 @@
           "steppedLine": false,
           "targets": [
             {
-              "expr": "node_load1{instance=~\"$server\"}",
+              "expr": "node_load1{instance=\"$server\"}",
               "intervalFactor": 4,
               "legendFormat": "load 1m",
               "refId": "A",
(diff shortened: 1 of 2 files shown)


assets/grafana/node-dashboard.json
```diff
@@ -405,15 +414,15 @@
           },
           "targets": [
             {
-              "expr": "(node_memory_MemFree{instance=~\"$server\"} / node_memory_MemTotal{instance=~\"$server\"}) * 100",
```
fabxc: This works?

I assume `$server` is the variable interpolated by Grafana. All regex matches are anchored in Prometheus though. So `=~` would be equivalent to `=` here unless `$server` contains regex chars itself, which it shouldn't though that's not escaped by anything.

If it's about allowing a port suffix, this should only work when `.+` or similar is added to the variable. Though I don't know when Grafana stops lexing the var here.

fabxc:
Looks good in general. Just a bit confused on that `$server` thing.

assets/grafana/node-dashboard.json
brancz: `$server` is indeed a grafana interpolated thing. Now that you mention it, I agree, it should all work with `=` just fine. I'll try.

brancz: Yep, I just replaced all `=~` with `=` and it works just as it did before.

fabxc: 👍

PR #47, opened 2017-02-21
dashboards: fix all nodes dashboard

brancz:
Same game as #46 .

@fabxc 

/cc @urzds


fabxc:
👍 

Can we get rid of `grafana-watcher` eventually in favor of http://docs.grafana.org/installation/configuration/#dashboards-json?

brancz:
It's not quite as easy as we have to redeploy grafana when a configuration changes as grafana does not watch file system notifications and reloads them. Additionally although there is this functionality for dashboards, it is lacking for datasources, so for the time being we'll have to stick with the `grafana-watcher`.

PR #51, opened 2017-03-01
manifests: don't skip apiserver cert verification

brancz:
As apiserver certs are usually correctly issued we can not skip verifying these, we just need to make sure we ask for the right `serverName`.

@fabxc


fabxc:
👍 awesome. Finally!

PR #52, opened 2017-03-01
remove grafana watcher

brancz:
https://github.com/coreos/prometheus-operator/pull/183 migrates the grafana-watcher to the Prometheus Operator repository so it can be removed from here.

@fabxc


fabxc:
👍

Issue #53, opened 2017-03-01
Evaluate used memory as "Total - MemFree - Buffers - Cached" and display stacked Used/Buffers/Cached/Free


rutsky:
Currently "Used" memory is evaluated as "Total - MemFree":
https://github.com/coreos/kube-prometheus/blob/615f45e/assets/grafana/node-dashboard.json#L308

During typical Linux server usage "free memory" value is very small: kernel caches all disk I/O and usually uses almost all free memory for different types of caches, so currently Grafana almost always will show that > 90 % of memory is "used", which from some point of view is true, but useless.

Much better to show in separate actually privately used by processes memory, size of buffers, size of caches and size of free memory:

```
          "targets": [
            {
              "expr": "node_memory_MemTotal{instance=\"$server\"} - node_memory_MemFree{instance=\"$server\"} - node_memory_Buffers{instance=\"$server\"} - node_memory_Cached{instance=\"$server\"}",
              "intervalFactor": 2,
              "refId": "A",
              "legendFormat": "memory used",
            },
            {
              "expr": "node_memory_Buffers{instance=\"$server\"}",
              "intervalFactor": 2,
              "refId": "B",
              "legendFormat": "memory buffers",
            },
            {
              "expr": "node_memory_Cached{instance=\"$server\"}",
              "intervalFactor": 2,
              "refId": "C",
              "legendFormat": "memory cache"
            },
            {
              "expr": "node_memory_MemFree{instance=\"$server\"}",
              "intervalFactor": 2,
              "refId": "D",
              "legendFormat": "memory free",
            }
          ],
```

This is done in prometheus dashboard by default: https://github.com/prometheus/prometheus/blob/bed4635/consoles/node-overview.html#L99

brancz:
Great! Thanks for noticing! Do you want to open a PR for this?

rutsky:
@brancz 
> Do you want to open a PR for this?

I'm trying, but looks like there are issues with updating dashboard manifests.

What is intended way of updating manifests in https://github.com/coreos/kube-prometheus/tree/master/assets/grafana ?

I though that it should be:

1) edit dashboard in live Grafana,
2) click "export",
3) update `assets/grafana/` manifests with exported jsons.
4) run https://github.com/coreos/kube-prometheus/blob/master/hack/scripts/generate-configmaps.sh and update configman in Kubernetes

[...]

PR #55, opened 2017-03-02
[DO NOT MERGE] split used memory into actually used memory, caches and buffers

rutsky:
Generated by editing dashboard in live Grafana, exporting to file,
wrapping file with as specified here:
https://github.com/coreos/kube-prometheus/issues/53#issuecomment-283603064

This PR is not intended for merging, but instead for viewing how changes made in live Grafana are looks like, see #53


rutsky:
Closing in favor of #56

PR #56, opened 2017-03-02
Split used memory on used, caches and buffers

### LINKED ISSUE #53: Evaluate used memory as "Total - MemFree - Buffers - Cached" and display stacked Used/Buffers/Cached/Free

rutsky:
Fixes #53 

Here how it looks now:

### DIFF
--- a/assets/grafana/all-nodes-dashboard.json
+++ b/assets/grafana/all-nodes-dashboard.json
@@ -1,5 +1,6 @@
 {
-  "dashboard": {
+  "dashboard":
+{
   "__inputs": [
     {
       "name": "DS_PROMETHEUS",
@@ -268,11 +269,11 @@
             }
           ],
           "span": 9,
-          "stack": false,
+          "stack": true,
           "steppedLine": false,
           "targets": [
             {
-              "expr": "sum(node_memory_MemTotal) - sum(node_memory_MemFree)",
+              "expr": "sum(node_memory_MemTotal) - sum(node_memory_MemFree) - sum(node_memory_Buffers) - sum(node_memory_Cached)",
               "intervalFactor": 2,
               "legendFormat": "memory usage",
               "metric": "memo",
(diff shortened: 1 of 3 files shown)


brancz:
Great stuff! Thanks a lot! Feel free to include the bash script you commented on in the `hack/scripts/` directory. It will definitely be useful in the future! :tada:

rutsky:
I added bash script here: #57 
Also I specified whole usage pipeline in it.

brancz:
good stuff! :+1: lgtm

PR #60, opened 2017-03-07
migrate kube-prometheus to Prometheus Operator repo

brancz:
With https://github.com/coreos/prometheus-operator/pull/192 merged, we can now remove it from here and leave a pointer to the new location.

@fabxc 

/cc fyi @mgoodness


fabxc:
:+1:

PR #67, opened 2019-04-15
Add missing adapter to build snippet

metalmatze:
Just noticed a difference between this file and the example.jsonnet.

Originally by @jgustie in https://github.com/coreos/prometheus-operator/pull/2556


brancz:
lgtm on green

PR #69, opened 2019-04-16
Remove node interface down

brancz:
This alert is too aggressive and alerts on interfaces that are expected to be down. I talked to the SDN team for OpenShift and they mentioned we'll need to go with another strategy but for now best is to remove this alert.

https://bugzilla.redhat.com/show_bug.cgi?id=1700057

@mxinden @metalmatze @squat @s-urbaniak @paulfantom


brancz:
(you can safely ignore the travis push job, I accidentally first pushed my branch onto the upstream repo, and then removed the branch before travis could clone it)

PR #80, opened 2019-04-23
jsonnet/prometheus-adapter: add metrics to default user-facing roles

s-urbaniak:
Currently, only cluster wide admins have the permissions to view metrics
resources. This fixes it by adding a read-only cluster role which includes
aggregation labels to synthesize permission rules for standard user-facing
roles according to [1].

Note that only the "pods" resource is granted as reading "nodes" metrics
requires a cluster wide permission.

[1]
https://kubernetes.io/docs/reference/access-authn-authz/rbac/#user-facing-roles

/cc @brancz @metalmatze

### DIFF
--- a/manifests/prometheus-adapter-clusterRoleAggregatedMetricsReader.yaml
+++ b/manifests/prometheus-adapter-clusterRoleAggregatedMetricsReader.yaml
@@ -1,0 +1,17 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-admin: "true"
+    rbac.authorization.k8s.io/aggregate-to-edit: "true"
+    rbac.authorization.k8s.io/aggregate-to-view: "true"
+  name: system:aggregated-metrics-reader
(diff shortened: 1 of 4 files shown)


brancz:
lgtm on green

manifests/prometheus-adapter-clusterRoleAggregatedMetricsReader.yaml
```diff
@@ -0,0 +1,17 @@
...
+  name: system:aggregated-metrics-reader
+rules:
+- apiGroups:
+  - metrics.k8s.io
+  resources:
+  - pods
+  verbs:
+  - get
+  - list
+  - watch
```
brancz: watch is not technically available for this API

s-urbaniak: This is technically true :+1: I am fine with removing it, leaving this here will make it readily available for the potential inclusion which is discussed upstream.

brancz: Agreed. Doesn't hurt. lgtm

PR #82, opened 2019-04-24
alerts: Replace ntp with timex metrics

brancz:
A local ntp server is optional, while timex is in kernel and always
present. Some ntp daemons like chrony can also only be run in a client
mode, which synchronizes the kernel but doesn't expose an ntp server.
Using the timex metrics is a more reliable approach.

@metalmatze @mxinden @squat @s-urbaniak @paulfantom


metalmatze:
LGTM
Interesting find. How did you find it?

s-urbaniak:
lgtm

PR #87, opened 2019-04-30
Use ksonnet.beta.4 with updated Kubernetes 1.14

metalmatze:
We are still blocked on https://github.com/ksonnet/ksonnet-lib/pull/160. I would say we give that PR a couple of more days and see if we can unblock this one. The we don't need to workaround by adding another dependency called `ksonnet4`.

/cc @squat @mxinden @paulfantom @brancz


brancz:
Nice!

bgagnon:
ksonnet/ksonnet-lib#160 is merged! looking forward to this!

brancz:
this needs an update but then should be ready to merge :tada:

metalmatze:
Moved to upstream `ksonnet.beta.4` and rebased against current master.

Issue #79, opened 2019-04-21
coredns servicemonitor change


marklester:
**What happened?**

prometheus doesn't pickup coredns target

**How to reproduce it (as minimally and precisely as possible)**:

install kube-prometheus with the quickstart instructions

**Environment**
* Prometheus Operator version:

quay.io/coreos/prometheus-operator:v0.29.0

* Kubernetes version information:

k8s version: v1.13.5

* Kubernetes cluster kind:

kubespray

* Manifests:

https://github.com/coreos/kube-prometheus/blob/master/manifests/prometheus-serviceMonitorCoreDNS.yaml#L19

**Anything else we need to know?**:
I believe that line should be:
``` 
k8s-app: coredns
```

brancz:
On most distributions the `Service` selecting CoreDNS pods, is still called `kube-dns` and is also labelled `k8s-app: kube-dns`. If this is different for kubespray, then the [kubespray jsonnet patch](https://github.com/coreos/kube-prometheus/blob/master/jsonnet/kube-prometheus/kube-prometheus-kubespray.libsonnet) should modify that.

Alan01252:
For anyone else looking at this, this is what I've done to generate working configuration with kubespray and kube-prometheus in my jsonnet file. I assume best practice would be to add this to the existing kubespray mixing instead?

```    
     _config+:: {
      namespace: 'monitoring',

      prometheus+:: {
      },

      jobs+: {
        CoreDNS: 'job="coredns"',
      },

    },

    prometheus+:: {

          serviceMonitorCoreDNS+: {
                spec+: {
                  selector: {
                        matchLabels: {
                          'k8s-app': 'coredns',
                        },
                  },
               },
          },


          serviceMonitorKubeScheduler+: {
                spec+: {
                  selector+: {
                        matchLabels: {
                          'k8s-app': 'kube-scheduler',
                        },
                  },
              },
          },

          serviceMonitorKubeControllerManager+: {
                spec+: {
                  selector+: {
                        matchLabels: {
                          'k8s-app': 'kube-controller-manager',
                        },
                  },
             },
          },

          kubeControllerManagerPrometheusDiscoveryService+: {
                spec+: {
                  selector: {
                        'component': 'kube-controller-manager',
                  },
              },
          },

           kubeSchedulerPrometheusDiscoveryService+: {
                spec+: {
                  selector: {
                    'component': 'kube-scheduler',
                  },
              },
          },

    },
`

[...]

PR #88, opened 2019-05-02
Update kubespray libsonnet for kubespray v2.9.0

Alan01252:
These changes have been tested against kubespray v2.9.0 on a baremetal
cluster created from scratch. They have not been tested against previous
versions of kubespray.

Fixes:

https://github.com/coreos/kube-prometheus/issues/79


Alan01252:
Actually, I'm assuming with this I should check in the manifest files that are generated which are specific for kubespray?

brancz:
I think you generated the `manifests/` directory from the kubespray patch? I think we should keep the manifests unopinionated, so just the jsonnet file changes sounds good to me.

Alan01252:
Manifests remove which makes sense to me.

brancz:
lgtm :+1: thanks for the contribution!

PR #89, opened 2019-05-03
Pin kubernetes-mixin v0.1.0

metalmatze:
We should use the released version of kubernetes-mixin. We can always release another kubernetes-mixin v0.1.x if we need anything updated.

### DIFF
--- a/jsonnet/kube-prometheus/jsonnetfile.json
+++ b/jsonnet/kube-prometheus/jsonnetfile.json
@@ -18,7 +18,7 @@
                     "subdir": ""
                 }
             },
-            "version": "master"
+            "version": "v0.1.0"
         },
         {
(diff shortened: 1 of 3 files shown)


metalmatze:
This will be even more important once we start merging Kubernetes 1.14 patches to the kubernetes-mixin `master` branch and thus would break people by updating the project's dependency here.

brancz:
We lock the exact version in the lock file already, here I would just “pin” against release-0.1, that’ll make updating a bit easier, especially for things that use the mixin as a transitive dependency.

metalmatze:
Updated the PR to pin `release-0.1`. Hope the CI works, as I haven't generate again.

brancz:
lgtm :+1:

PR #92, opened 2019-05-08
Note prometheus adapter component

StevenACoffman:
Per coreos/prometheus-operator#2519

I think it could better be advertised that you should not need to install metrics-server AND the prometheus operator.


metalmatze:
Yes, we can update the kube-prometheus docs on this. Happy to merge and addition to the README.md.

Just pointing out that you can install metrics-server AND the prometheus-operator, just the kube-prometheus stack ships the Prometheus Adapter instead of the metrics server. So that combination doesn't work. (I think you meant that, just making it more clear to others)

PR #94, opened 2019-05-09
update prometheus cluster role so it can scrape custom namespaces

huangqun:


metalmatze:
The file you edited is a generated file. You need to update this in the jsonnet source and then run `make generate` after commit the initial change:
https://github.com/coreos/kube-prometheus/blob/12724e2be0b228c75b2f66afa0c75e085a1c2fc3/jsonnet/kube-prometheus/prometheus/prometheus.libsonnet#L74-L91

I'm not entirely sure if we really need that change though. Can you give a bit more background to why you want this?

huangqun:
Sorry I wasn't aware it's a generated file. We found this necessary because we have a lot of projects in our cluster and each project has its own namespace, without these changes Prometheus is not able to scrape data from pods running under these custom namespaces.

Is there a better way to do this other than changing the prometheus cluster role?

Thanks!

brancz:
This is intentionally not here for least privilege reasons. If you want further namespaces to be monitored, they need to be whitelisted, and then the role and rolebindings will be automatically generated. Should you truly want this ClusterRole then I'd ask you to do this customization in your downstream usage of the jsonnet kube-prometheus library.

huangqun:
Makes sense, closing this.

PR #101, opened 2019-05-16
*: cut v0.1

metalmatze:
Before cutting the release I bumped the dependencies to use Prometheus Operator v0.30.0.

We still need to add that sentence on how to use the release. Could you give me a rough idea @brancz, then I'll add that and we can release.

/cc @paulfantom @s-urbaniak @squat


brancz:
Super awesome! lgtm :+1:

PR #118, opened 2019-06-07
Add commented imports for most common patches to example.jsonnet

metalmatze:
I think we can improve the entry point for new people, by adding the most common imports to `example.jsonnet` so that those only need to be uncommented. 

/cc @brancz @paulfantom @squat @s-urbaniak


brancz:
Very nice idea!

PR #119, opened 2019-06-12
Update Prometheus, Node Exporter & Grafana

rajatvig:


jsonnet/kube-prometheus/prometheus/prometheus.libsonnet
```diff
@@ -6,7 +6,7 @@ local k3 = import 'ksonnet/ksonnet.beta.3/k.libsonnet';
     namespace: 'default',
 
     versions+:: {
-      prometheus: 'v2.7.2',
+      prometheus: 'v2.9.2',
```
brancz: v2.10.0 is already out?

rajatvig: Will update and run the regenerate

rajatvig:
@brancz Have updated the generated manifests. There is an issue though as the repository depends on itself, the lock file changes which I have omitted. It would likely need a bump once this merges to master.

brancz:
lgtm thanks!

PR #122, opened 2019-06-13
Fix Thanos deployment and service for ksonnet.beta.4

rajatvig:
It has been since the upgrade to ksonnet.beta.4. I didn't find any default generation for it to update.


brancz:
cc @metalmatze

Issue #72, opened 2019-04-16
High CPU throttling on kube-rbac-proxy containers with default installation


bgagnon:
Using a reasonably small K8S cluster:
* 5 etcd nodes
* 3 master nodes
* 5 worker nodes

And deploying `kube-prometheus` using the jsonnet-bundler method at `master`, I immediately get alerts firing such as:

> 56% throttling of CPU in namespace monitoring for container `kube-rbac-proxy-self` in pod `kube-state-metrics-5bd984db7b-m66m6`.

> 67% throttling of CPU in namespace monitoring for container `kube-rbac-proxy-main` in pod `kube-state-metrics-5bd984db7b-m66m6`.

> 53% throttling of CPU in namespace runway-monitoring for container `kube-rbac-proxy` in pod `node-exporter-7jlrq`.

The node-exporter alert is fired for all nodes (ie. all pods within the daemonset), so 13 times in my case.

**Source code**

The jsonnet libs in this repo have the following:

*node-exporter*
```js
        container.mixin.resources.withLimits({ cpu: '20m', memory: '40Mi' }) +
```

*kube-state-metrics*
```js
        container.mixin.resources.withLimits({ cpu: '20m', memory: '40Mi' });
```

**Questions**

* is this throttling alert _only_ caused by going over container cgroup limits or could this be external (ie. hypervisor side effects)
* are the containers behaving normally (ie. not doing more work than they should)
* is the `20m` limit realistic in real world scenarios?
* the values are not configurable through `_config`; can they be overridden nonetheless?

sichvoge:
You should be able to configure it at least. See: https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/master/config.libsonnet#L66

metalmatze:
I have the feeling that we increased these limits in the past, not sure what happened :eyes: 
We should probably just increase the defaults on those containers, rather than patching the alert itself.

With such low limits the container is hold back from its CPU time a lot by the Linux cfs scheduler (https://en.wikipedia.org/wiki/Completely_Fair_Scheduler).

brancz:
All these values are just suggestions, we're unlikely to find a good default for everyone, I suggest to remove the limits from everything, look at the actual usage, and pin limits on around that + 20%.

goll:
FWIW after having the same issues ended up removing the limits completely as OpenShift did: https://github.com/openshift/cluster-monitoring-operator/pull/19

brancz:
[...]

PR #132, opened 2019-06-24
Add mixin to strip cpu/memory limits (issue #72)

bgagnon:
This mixin will strip the `limits` fields of certain containers with very low CPU limits which triggers throttling alerts (see #72). These are quite possibly due to a Kernel bug with CFS and mostly harmless, but the noise caused by the false alarms is undesirable enough to warrant this patch.

Patched containers:
- `addon-resizer`
- `kube-rbac-proxy`
- `kube-rbac-proxy-main`
- `kube-rbac-proxy-self`
- `node-exporter`

To strip the limits on the config reload containers added as sidecars to Prometheus pods, the CLI flag `--config-reloader-cpu=0` is added to the args of the `prometheus-operator` container.

### DIFF
--- a/jsonnet/kube-prometheus/kube-prometheus-strip-limits.libsonnet
+++ b/jsonnet/kube-prometheus/kube-prometheus-strip-limits.libsonnet
@@ -7,10 +7,10 @@
         template+: {
           spec+: {
             local stripLimits(c) =
-                if std.count([
-                    'node-exporter',
-                    'kube-rbac-proxy'
-                ], c.name) > 0
+                if std.setMember(c.name, [
+                  'kube-rbac-proxy',
+                  'node-exporter',
+                ])
                 then c + {resources+: {limits: {}}}
                 else c,
             containers: std.map(stripLimits, super.containers),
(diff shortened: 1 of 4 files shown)


jsonnet/kube-prometheus/kube-prometheus-strip-limits.libsonnet
```diff
@@ -0,0 +1,56 @@
+// Strips spec.containers[].limits for certain containers
+// https://github.com/coreos/kube-prometheus/issues/72
+{
```
brancz: contained within this mixin I think we can have a config object for each container's resources

bgagnon: Sure! Something like this?

```jsonnet
  _config+:: {
    nodeExporter+: {
      containerLimits+: {
        kubeRbacProxy+: {
          cpu: 0,
          memory: 0,
        },
        nodeExporter+: {
          cpu: 0,
          memory: 0,
        },
      },
    },
    prometheusOperator+: {
      containerLimits: {
        configReloader: {
          cpu: 0,
          memory: 0,
        },
      },
    },
    kubeStateMetrics+: {
      containerLimits: {
        addonResizer: {
          cpu: 0,
          memory: 0,
        },
        kubeRbacProxyMain: {
          cpu: 0,
          memory: 0,
        },
        kubeRbacProxySelf: {
          cpu: 0,
          memory: 0,
        },
      },
    },
  },
```

Or perhaps all the RBAC proxies can be configured in one place?

brancz: > Or perhaps all the RBAC proxies can be configured in one place?

Yes I think that would be appropriate.

I was more imagining something along the lines of this for the configuration:

```jsonnet
  _config+:: {
    resources+:: {
      "node-exporter": {}, // container name as key, so we can just set the resources to what matches here, but by "default" set all of them to empty so by default the patch removes all resource requests and limits
      "kube-rbac-proxy": {
        requests: {cpu: "100m", "50Mi"}, // example demonstration of how we would set requests
      },
    },
  },
```

What do you think?

PR #133, opened 2019-06-25
Use kubernetes-mixin's master in kube-prometheus master

metalmatze:
It seems, that somehow we've moved to using `release-0.1` on kube-prometheus `master` too. This should be kubernetes-mixin's `master` here, as both `master` branches work for Kubernetes 1.14+.

/cc @brancz @paulfantom @squat


brancz:
Oops! Still need to actually bump this again in cmo.

PR #140, opened 2019-07-05
Add thanos v0.5.0 support & remove deprecated thanos gossip

benjaminhuo:
Signed-off-by: Benjamin <benjamin@yunify.com>

### DIFF
--- a/jsonnet/kube-prometheus/kube-prometheus-thanos.libsonnet
+++ b/jsonnet/kube-prometheus/kube-prometheus-thanos.libsonnet
@@ -5,7 +5,7 @@ local servicePort = k.core.v1.service.mixin.spec.portsType;
 {
   _config+:: {
     versions+:: {
-      thanos: 'v0.3.2',
+      thanos: 'v0.5.0',
     },
     imageRepos+:: {
       thanos: 'improbable/thanos',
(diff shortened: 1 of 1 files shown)


brancz:
lgtm :+1: thanks a lot for taking care of this! :)

benjaminhuo:
:) I planed to submit a PR for the prometheus operator to remove thanos gossip, however it's been done when I have time to do that. So I can help on this part ...

Btw, I've noticed that https://github.com/ksonnet/ksonnet will be archived , and this project is
 still using https://github.com/ksonnet/ksonnet-lib, so will this be changed in the future?

Another question is that the CompatibilityMatrix of prometheus operator supports up to  prometheus 2.7.1, while kube-prometheus' prometheus has been updated to 2.10. So will this new version work fine with prometheus operator?

Thanks
Ben

brancz:
> Btw, I've noticed that https://github.com/ksonnet/ksonnet will be archived , and this project is
> still using https://github.com/ksonnet/ksonnet-lib, so will this be changed in the future?

We've started contributing to this (for example we contributed the latest 1.14 support to ksonnet-lib). Should they actually archive it then we'll take over maintenance somewhere else :slightly_smiling_face: . (note though we've spoken to the maintainers and they've hinted that there is no such plan for the lib yet)

> Another question is that the CompatibilityMatrix of prometheus operator supports up to prometheus 2.7.1, while kube-prometheus' prometheus has been updated to 2.10. So will this new version work fine with prometheus operator?

We should add 2.10 to the matrix we just haven't, it should work perfectly fine, but it's worth adding because all the versions in the compatibility matrix we actually test in every CI run :)

benjaminhuo:
> We should add 2.10 to the matrix we just haven't, it should work perfectly fine, but it's worth adding because all the versions in the compatibility matrix we actually test in every CI run :)

Then can I submit a PR to the prometheus operator to add 2.7.2, 2.8.1, 2.9.2, 2.10.0 to CompatibilityMatrix ?

brancz:
I’d be ok with keeping the 1.x releases we have in there already, but would be happy to have the new ones! :)

benjaminhuo:
> I’d be ok with keeping the 1.x releases we have in there already, but would be happy to have the new ones! :)

Got it

curantes:
Any chance this PR could also be merged to release-1.0 branch. My clusters are still att Kuberntes v1.13.x and I would like to use newer Thanos without the gossip.

brancz:
@curantes we're unlikely to do that, but happy to accept PRs :)

PR #141, opened 2019-07-05
Increase ntp clockskewdetected accuracy to 5ms

dominikhahn:


metalmatze:
Why is 5ms better than 3ms for you? :blush:

caarlos0:
I don't know if 5ms is good, but I would say this could be higher... this alert fires all the time even on a new GKE cluster... it has became mostly noise...

brancz:
Funnily enough I keep hearing this only about GKE ... but I think 5ms is just as good of a default. Let's go with it and adapt as needed.

benclapp:
FWIW we've seen this firing frequently (at 3ms) on IKS and EKS. EKS is more prevalent in some regions over others.

PR #144, opened 2019-07-09
Update kube-state-metrics.libsonnet

zachaller:
With the bump to kube-state-metrics v1.6 they added ingress and certificates monitoring this updates the rbac rules so that those work with the new version.


brancz:
Thanks! lgtm :+1:

PR #147, opened 2019-07-10
Fix the Thanos Query deployment to use Thanos Store service

rajatvig:
1. Thanos Query deployment is bound to 10902 on HTTP not 9090
2. Thanos Query needs Thanos Store address, not Prometheus address
3. Thanos Sidecar Service should really be the Thanos Store Service

### DIFF
--- a/jsonnet/kube-prometheus/kube-prometheus-thanos.libsonnet
+++ b/jsonnet/kube-prometheus/kube-prometheus-thanos.libsonnet
@@ -43,7 +43,7 @@ local servicePort = k.core.v1.service.mixin.spec.portsType;
           '--log.level=debug',
           '--query.replica-label=prometheus_replica',
           '--query.auto-downsampling',
-          '--store=dnssrv+thanos-sidecar.' + $._config.namespace + '.svc',
+          '--store=dnssrv+thanos-store.' + $._config.namespace + '.svc:10901',
         ]);
       local podLabels = { app: 'thanos-query' };
       deployment.new('thanos-query', 1, thanosQueryContainer, podLabels) +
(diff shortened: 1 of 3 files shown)


jsonnet/kube-prometheus/kube-prometheus-thanos.libsonnet
```diff
@@ -43,7 +43,7 @@ local servicePort = k.core.v1.service.mixin.spec.portsType;
           '--log.level=debug',
           '--query.replica-label=prometheus_replica',
           '--query.auto-downsampling',
-          '--store=dnssrv+thanos-sidecar.' + $._config.namespace + '.svc',
+          '--store=dnssrv+thanos-store.' + $._config.namespace + '.svc:10901',
```
brancz: we need both sidecar and store here no? cc @metalmatze

rajatvig: The reference is to the DNS based SD for finding the store service. The sidecar service was acting as the proxy to Prometheus in the last update. It should instead point to the Store service.

It was also missing the GRPC port of the store service.

brancz: I mean we want to discover both sidecars as well as stores, so both flags, instead of removing the sidecar one:

```
          '--store=dnssrv+thanos-sidecar.' + $._config.namespace + '.svc',
          '--store=dnssrv+thanos-store.' + $._config.namespace + '.svc:10901',
```

metalmatze: Yes, we want both. Additionally, we also want to use the proper way of using dnssrv records.
Maybe you could update that in this PR too?

https://github.com/metalmatze/kube-thanos/blob/master/manifests/thanos-querier-deployment.yaml#L22

metalmatze:
Sorry, this might be conflicting / be outdated by #149.
We have started working on a separate project called [kube-thanos](github.com/metalmatze/kube-thanos) which is focused specifically on Thanos. Only the sidecar should survive in kube-prometheus.

rajatvig:
@metalmatze That is great. Have some home grown rules for Thanos that I was planning to add next. The new repository is a much better place for it.

PR #150, opened 2019-07-11
Use upstream Prometheus monitoring mixin

brancz:
This removes the home-grown alerting rules with the recently merged Prometheus monitoring mixin.

@metalmatze @kakkoyun @paulfantom


metalmatze:
Once https://github.com/prometheus/prometheus/pull/5753 gets merged we can update this PR or just do a follow up to get Prometheus dashboards from upstream too! :tada:

PR #157, opened 2019-07-17
change to use GA kubelet OS label

rphillips:
change to use the GA kubelet OS label

ref: https://github.com/openshift/cluster-monitoring-operator/issues/371
ref: https://jira.coreos.com/browse/NODE-114

### DIFF
--- a/jsonnet/kube-prometheus/alertmanager/alertmanager.libsonnet
+++ b/jsonnet/kube-prometheus/alertmanager/alertmanager.libsonnet
@@ -112,7 +112,7 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
           replicas: $._config.alertmanager.replicas,
           version: $._config.versions.alertmanager,
           baseImage: $._config.imageRepos.alertmanager,
-          nodeSelector: { 'beta.kubernetes.io/os': 'linux' },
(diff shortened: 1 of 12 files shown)


squat:
It seems that you manually changed the node selectors in the generated manifests. In order for the generation to work, we need to bump the hash in https://github.com/coreos/kube-prometheus/blob/master/jsonnetfile.lock.json#L11

This is because the example manifests vendor kube-prometheus itself. After updating the hash to the hash of you commit that introduces the libsonnet changes, regenerate and commit that result in a “regenerate” commit. Making jsonnet changes to the repo itself is currently always a two-step process, but this is largely a limitation of jb today and will be improved soon.

rphillips:
@squat thanks for the tip! updated the PR and regenerated.

rphillips:
`wait for prometheus-k8s: timed out waiting for the condition`...

/retest

rphillips:
Strange... same error. Thoughts?

brancz:
Is kube 1.13 maybe not new enough?

rphillips:
Kube 1.14 is needed for this feature.

brancz:
I see kind node v1.13 in the build logs. Can you upgrade this?

brancz:
Awesome! lgtm :+1: thanks!

PR #162, opened 2019-07-24
Include node name in NodeDiskRunningFull alert

LiliC:
This relabels the instance label to node name value, instead of IP. This PR also includes the node name in the `NodeDiskRunningFull` alert message. This makes it easier to identify which node alert is firing for. Currently, only device namespace and pod name were included in
the alert.

cc @s-urbaniak @paulfantom @brancz 

Alert:

Screenshot 2019-07-24 at 17 58 38


brancz:
Very nice! lgtm :+1:

Issue #161, opened 2019-07-23
minikube pods in pending state


karancode:
**What did you do?**
I followed the Quickstart steps [here](https://github.com/coreos/kube-prometheus#minikube) to get kube-prometheus running on my minikube but all the pods are in pending state (except grafana & prometheus-operator).
Am I missing something ?

**Did you expect to see some different?**
Pods should be in Running State.

**Environment**
OS : macOS Mojave version: 10.14.5
minikube version: v1.1.0

* Prometheus Operator version:

    `quay.io/coreos/prometheus-operator:v0.31.1`

* Kubernetes version information:
 ```
Client Version: version.Info{Major:"1", Minor:"13", GitVersion:"v1.13.0", GitCommit:"ddf47ac13c1a9483ea035a79cd7c10005ff21a6d", GitTreeState:"clean", BuildDate:"2018-12-03T21:04:45Z", GoVersion:"go1.11.2", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"13", GitVersion:"v1.13.5", GitCommit:"2166946f41b36dea2c4626f90a77706f426cdea2", GitTreeState:"clean", BuildDate:"2019-03-25T15:19:22Z", GoVersion:"go1.11.5", Compiler:"gc", Platform:"linux/amd64"}
```

* Kubernetes cluster kind:
verifying on `minikube`.

* Manifests:
Using from the latest commit 16a49f00d6c77ad85abbfdba49e186808daf2764

* Prometheus Operator Logs:
No error logs from the prometheus-operator pod.
pasted logs [here](https://pastebin.com/n7aR0hE6)

* Prometheus Logs:
No logs for prometheus

**Anything else we need to know?**:
N/A

**Comment by brancz** [comment]
Did you start minikube using the command in the readme? What do events say?

**Comment by karancode** [comment]
@brancz Yes. I think minikube start was ok.
```
$ minikube delete && minikube start --kubernetes-version=v1.13.5 --memory=4096 --bootstrapper=kubeadm --extra-config=kubelet.authentication-token-webhook=true --extra-config=kubelet.authorization-mode=Webhook --extra-config=scheduler.address=0.0.0.0 --extra-config=controller-manager.address=0.0.0.0
There is a newer version of minikube available (v1.2.0).  Download it here:
https://github.com/kubernetes/minikube/releases/tag/v1.2.0

To disable this notification, run the following:
minikube config set WantUpdateNotification false
[...]

PR #168, opened 2019-07-25
update README, change minikube k8s-version to 1.14.4

karancode:
As @brancz shared in issue #161, for latest changes we need k8s-version 1.14+
This is to update the README in master.


brancz:
Thanks a lot! :+1:

PR #175, opened 2019-08-02
Fix selectors for Kubernetes 1.14+

sturman:
According to [comment](https://github.com/coreos/kube-prometheus/issues/161#issuecomment-514432211) in issue #161 for Kubernetes 1.14+ selector `kubernetes.io/os` should be used instead `beta.kubernetes.io/os`

### DIFF
--- a/manifests/0prometheus-operator-deployment.yaml
+++ b/manifests/0prometheus-operator-deployment.yaml
@@ -41,7 +41,7 @@ spec:
         securityContext:
           allowPrivilegeEscalation: false
       nodeSelector:
-        beta.kubernetes.io/os: linux
+        kubernetes.io/os: linux
       securityContext:
         runAsNonRoot: true
         runAsUser: 65534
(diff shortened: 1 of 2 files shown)


brancz:
These manifests are generated from the jsonnet files in the prometheus operator repo, hence it first needs to be updated there: https://github.com/coreos/prometheus-operator/blob/cc584ecfa08d2eb95ba9401f116e3a20bf71be8b/jsonnet/prometheus-operator/prometheus-operator.libsonnet#L151

metalmatze:
Take a look at https://github.com/coreos/kube-prometheus/pull/157/files as reference with similar changes.

BenoitKnecht:
Both instances of `beta.kubernetes.io/os` addressed in this PR come from external dependencies, namely:

- https://github.com/brancz/kubernetes-grafana/blob/master/grafana/grafana.libsonnet#L176
- https://github.com/coreos/prometheus-operator/blob/master/jsonnet/prometheus-operator/prometheus-operator.libsonnet#L164

Given that the label [has been deprecated in 1.14 and is scheduled for removal in 1.18](https://kubernetes.io/docs/setup/release/notes/#deprecations-and-removals), it will be migrated when these projects stop supporting k8s < 1.14.

Therefore, I think this PR should be closed.

PR #178, opened 2019-08-08
Drop container_* metrics with no image.

taizoKaneko:
some metrics related to container is collected twice.
My issue:
https://github.com/helm/charts/issues/16121
Related issue:
https://github.com/kubernetes-monitoring/kubernetes-mixin/issues/136

Despite [the solution is known](https://github.com/kubernetes-monitoring/kubernetes-mixin/issues/136#issuecomment-456033706), no one seems to send a pull request.

### DIFF
--- a/jsonnet/kube-prometheus/prometheus/prometheus.libsonnet
+++ b/jsonnet/kube-prometheus/prometheus/prometheus.libsonnet
@@ -285,6 +285,12 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
               },
               bearerTokenFile: '/var/run/secrets/kubernetes.io/serviceaccount/token',
               metricRelabelings: [
+                // Drop container_* metrics with no image.
+                {
+                  sourceLabels: ['__name__', 'image'],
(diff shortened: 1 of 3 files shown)


taizoKaneko:
I only committed my part, should I include this link too?
https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/232

brancz:
I don’t think these metrics should be dropped. It’s intentional that the entire cgroups hierarchy is exposed, this should be handled at query time.

taizoKaneko:
I understood and agree.
So, what should be fixed is grafana, is this correct?
(if focusing on container_memory_rss, like)
https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/master/dashboards/resources.libsonnet#L98

brancz:
Correct

taizoKaneko:
Thank you, @brancz.

taizoKaneko:
Sorry, I was wrong.
Those metrics were not collected twice. Everything is fine.

But I don't know why I had a problem before.
If this happens again, I will work on it.

PR #180, opened 2019-08-09
Fix io time yaxe

lord63:
The upstream has be updated and fixed kubernetes-monitoring/kubernetes-mixin#235 , so the new generated manifests should be good, but in case that someone will use the generated manifests directly, I guess it would be better to update the manifests?

### DIFF
--- a/jsonnetfile.lock.json
+++ b/jsonnetfile.lock.json
@@ -28,7 +28,7 @@
                     "subdir": ""
                 }
             },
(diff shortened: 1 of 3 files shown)


lord63:
Hmmm, will fix the test, hold on.

lord63:
Tests passed, commits squashed, should be good to go.

brancz:
Thanks! lgtm

PR #182, opened 2019-08-09
set kubeStateMetrics version to 1.7.2

guusvw:
to avoid using the vulnerable version of `kube-state-metrics`

https://github.com/kubernetes/kube-state-metrics/releases/tag/v1.7.2


brancz:
Thanks!

PR #185, opened 2019-08-12
Fix wrong config used for node-exporter container limits

bgagnon:
Simple mistake that slipped through during #132.


metalmatze:
Thanks!

PR #184, opened 2019-08-12
Use local dependency for kube-prometheus jsonnet

metalmatze:
Once we merge https://github.com/jsonnet-bundler/jsonnet-bundler/pull/36 we can make use of local dependencies with jsonnet-bundler.
Before the build for this PR in the CI succeeds, we need to update `jb` in the CI image (/cc @paulfantom).

/cc @brancz @LiliC @s-urbaniak @squat

PR #186, opened 2019-08-15
*: Fix AlertmanagerConfigInconsistent many-to-many matching errors

LiliC:
If there is more than one prometheus-operator pod, which happens briefly
when we delete the prometheus-operator pod, we can see the errors of
many-to-many matching, this whitelists the labels matching, and excluded
the pod.

cc @metalmatze


LiliC:
The `prometheus-k8s` is not available it seems, according to the e2e failing test. Will try to reproduce locally. But if anyone has any hints, would be appreciated.

brancz:
So sorry, I think I caused this conflict by merging #184. Because I merged #184 this should never happen again because of this reason! :tada:

brancz:
Awesomeeeeeee :tada: lgtm :+1:

PR #189, opened 2019-08-21
Add labels for namespace and service to TargetDown

kylebyerly-hp:
Idea from here https://github.com/alphagov/gsp/pull/328

This enables alerts for down targets to have fields such as namespace and service

### DIFF
--- a/jsonnet/kube-prometheus/alerts/general.libsonnet
+++ b/jsonnet/kube-prometheus/alerts/general.libsonnet
@@ -9,7 +9,7 @@
             annotations: {
               message: '{{ $value }}% of the {{ $labels.job }} targets are down.',
             },
-            expr: '100 * (count(up == 0) BY (job) / count(up) BY (job)) > 10',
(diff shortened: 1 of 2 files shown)


jsonnet/kube-prometheus/alerts/general.libsonnet
```diff
@@ -9,7 +9,7 @@
             annotations: {
               message: '{{ $value }}% of the {{ $labels.job }} targets are down.',
             },
-            expr: '100 * (count(up == 0) BY (job) / count(up) BY (job)) > 10',
+            expr: '100 * (avg without(instance, pod) (up)) < 90',
```
brancz: I'm a little wary of this, as we have no idea what people might be doing with their target's labels. I think I'd rather do:

```
100 * (count(up == 0) BY (job, namespace, service) / count(up) BY (job, namespace, service)) > 10
```

I think those are the most likely to be grouping labels. What do you think?

brancz:
Thanks!

PR #195, opened 2019-08-27
use real alertmanager name

prune998:
`alertmanager-main` is hardcoded where it should derive from the real alertmanager name as defined by the user

This also apply to the release branch...


brancz:
Do you want to open a PR against the release branch as well?

PR #203, opened 2019-08-30
Fix unmatched, misplaced braces

oneandwholly:
I noticed some components in the dashboard were broken, namely, Kubernetes / StatefulSets 
and Disk Space Utilisation at USE Method / Node. I don't know if the person who wrote it left the trailing commas on purpose, but removing it made it work.

### DIFF
--- a/manifests/grafana-dashboardDefinitions.yaml
+++ b/manifests/grafana-dashboardDefinitions.yaml
@@ -19485,7 +19485,7 @@ items:
                       "options": [
 
                       ],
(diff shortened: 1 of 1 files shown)


oneandwholly:
i guess i have to generate these files instead of fixing them manually

brancz:
> i guess i have to generate these files instead of fixing them manually

that is correct, and beyond, these dashboards are defined in the [kubernetes-mixin](https://github.com/kubernetes-monitoring/kubernetes-mixin) project, so they need to first be fixed there :)

PR #212, opened 2019-09-07
Update README.md

stillzhl:
The `example.jsonnet` is out of date for thanos libsonnet. I got blow error when I uncommented thanos libsonnet import:
```
RUNTIME ERROR: couldn't open import "kube-prometheus/kube-prometheus-thanos-sidecar.libsonnet": no match locally or in the Jsonnet library paths
	my-kube-prom.jsonnet:8:4-69	thunk <kp> from <$>
	my-kube-prom.jsonnet:15:81-83
	<std>:1268:24-25	thunk from <function <anonymous>>
	<std>:1268:5-33	function <anonymous>
	my-kube-prom.jsonnet:15:64-99	$
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,7 +178,7 @@ local kp =
   // (import 'kube-prometheus/kube-prometheus-managed-cluster.libsonnet') +
   // (import 'kube-prometheus/kube-prometheus-node-ports.libsonnet') +
   // (import 'kube-prometheus/kube-prometheus-static-etcd.libsonnet') +
-  // (import 'kube-prometheus/kube-prometheus-thanos-sidecar.libsonnet') +
+  // (import 'kube-prometheus/kube-prometheus-thanos.libsonnet') +
   {
     _config+:: {
       namespace: 'monitoring',


brancz:
The actual `example.jsonnet` file is embedded into the readme. Could you also update that file? Thanks! :slightly_smiling_face:

BenoitKnecht:
@stillzhl I think there's been some confusion here.

In `master`, `README.md` refers to `kube-prometheus/kube-prometheus-thanos-sidecar.libsonnet`, and that file does exist.

In `release-0.1`, `kube-prometheus/kube-prometheus-thanos-sidecar.libsonnet` doesn't exist, but there's a file called `kube-prometheus/kube-prometheus-thanos.libsonnet` instead; however, neither `README.md` nor `example.jsonnet` refer to either of those files in that branch.

My guess is you were looking at `README.md` on GitHub (i.e. the `master` version), but you had `release-0.1` checked out locally, hence the mix up.

So I think this PR should be closed since both `release-0.1` and `master` are consistent.

PR #215, opened 2019-09-09
jsonnet/prometheus-adapter: use pod cgroup for for cpu/memory metrics.

s-urbaniak:
We have to use container_name="" and pod_name!="" to get memory/cpu usage
that is accounted to the pod cgroup.

xref https://github.com/openshift/console/pull/2486#discussion_r317647400

/cc @paulfantom @LiliC @metalmatze @brancz

### DIFF
--- a/jsonnet/kube-prometheus/prometheus-adapter/prometheus-adapter.libsonnet
+++ b/jsonnet/kube-prometheus/prometheus-adapter/prometheus-adapter.libsonnet
@@ -19,7 +19,7 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
       config: |||
         resourceRules:
           cpu:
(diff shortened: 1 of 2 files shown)


s-urbaniak:
fwiw: I tested this locally on a kind cluster.

brancz:
This should be on the container level no? Otherwise I don't see how:

```
kubectl top pod --containers
```

Would work.

brancz:
(on a different note this might mean that the adapter might need an additional configuration to configure a query for each: node, pod, container)

s-urbaniak:
@brancz indeed I didn't verify `--containers`, but you are correct. This will need another configuration option for a per-container query.

s-urbaniak:
closing out this PR as it requires changes in prometheus adapter.

PR #226, opened 2019-09-17
Fix some Kubernetes rules displaying percent but being ratio (multipl…y them by 100)

desaintmartin:
BTW, I did not found any upstream, shouldn't this be in kubernetes-mixin?

### DIFF
--- a/manifests/prometheus-rules.yaml
+++ b/manifests/prometheus-rules.yaml
@@ -242,7 +242,7 @@ spec:
         summary: Filesystem is predicted to run out of space within the next 24 hours.
       expr: |
         (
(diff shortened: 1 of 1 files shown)


brancz:
These rules are defined in the node-mixin any changes first have to be made there, then the dependency can be updated here

https://github.com/prometheus/node_exporter/tree/master/docs/node-mixin

BenoitKnecht:
This has been fixed in prometheus/node_exporter@dfeec07f2fcb07f49d91108861b98c4a859f93a8, and kube-prometheus now depends on a version that includes the fix, so this PR can be closed.

brancz:
Thanks a lot @BenoitKnecht! Closing.

PR #230, opened 2019-09-19
adding note for k8s before v1.14.0

dparkar:
This will help for folks who want to deploy on Azure Kubernetes Service where the latest version is 1.12.8 as of 9/19/19.


brancz:
Thanks!

PR #233, opened 2019-09-23
jsonnet/kube-prometheus/kube-state-metrics: Drop ksm own labels

LiliC:
This PR drops the labels specific to kube-state-metrics pod from kube-state-metrics metrics. These labels are misleading as they are about the kube-state-metrics pod, now about the resources themselves.

Before:
```
kube_replicaset_metadata_generation{endpoint=“https-main”,instance=“1234:8443",job=“kube-state-metrics”,namespace=“cert-manager”,pod=“kube-state-metrics-64c4bdf979-tlzjr”,replicaset=“cert-manager-b9748ff89",service=“kube-state-metrics”}
```
After:
```
kube_replicaset_metadata_generation{instance=“1234:8443”,namespace=“kube-system”,replicaset=“coredns-584795fc57”}
```

Question: Not sure if the instance label is about the ksm pod node or the resource node?

cc @brancz

### DIFF
--- a/jsonnet/kube-prometheus/kube-state-metrics/kube-state-metrics.libsonnet
+++ b/jsonnet/kube-prometheus/kube-state-metrics/kube-state-metrics.libsonnet
@@ -310,6 +310,12 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
               scrapeTimeout: $._config.kubeStateMetrics.scrapeTimeout,
               honorLabels: true,
               bearerTokenFile: '/var/run/secrets/kubernetes.io/serviceaccount/token',
+              relabelings: [
+                {
+                  regex: '(pod|service|job|endpoint|namespace)',
+                  action: 'labeldrop',
+                },
+              ],
               tlsConfig: {
                 insecureSkipVerify: true,
               },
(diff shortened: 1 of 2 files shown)


jsonnet/kube-prometheus/kube-state-metrics/kube-state-metrics.libsonnet
```diff
@@ -310,6 +310,12 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
               scrapeTimeout: $._config.kubeStateMetrics.scrapeTimeout,
               honorLabels: true,
               bearerTokenFile: '/var/run/secrets/kubernetes.io/serviceaccount/token',
+              relabelings: [
+                {
+                  regex: '(pod|service|job|endpoint)',
```
brancz: we want to also drop namespace here, no?

LiliC: No, thats the correct namespace, see in the example.

LiliC:
@brancz done PTAL, thanks!

brancz:
lgtm on green

wu0407:
The "instance" label is  ksm pod ip and port, it can remove.
but "pod" and "namespace" should not remove， like kube_pod_info、kube_pod_labels  identity by this label

brancz:
This is only on the target, not the metrics :) the kube_pod_info etc. metrics will still have namespace and pod labels

PR #234, opened 2019-09-24
Fix KubeMismatchVersion false positive alert on Kubernetes < 1.14

minhdanh:
This is a continuity of https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/253
to fix fix false positive kubeversionmismatch for Kubernetes < 1.14
After this I'll create a PR to sync the rules in prometheus-opeartor chart.


brancz:
Thanks!

PR #245, opened 2019-09-26
jsonnet/kube-prometheus/kube-state-metrics: Do not drop job label

LiliC:
This broke some alerts, as they are using the `job="kube-state-metrics"` label. We should first remove that label from the alerts and then drop this label. For now, lets "revert" the job label dropping.

cc @brancz


brancz:
lgtm on green

PR #246, opened 2019-09-26
fix apiVersion

DesistDaydream:
In the kubernetes 1.16 release, the apiVersion of the deployment has changed.


LiliC:
Thanks for this, but this was already fixed in the upstream dependency. These files are generated. The fix for this is in https://github.com/coreos/kube-prometheus/pull/247

PR #248, opened 2019-09-27
jsonnet/prometheus-adapter: Fix query for k8s 1.16

s-urbaniak:
/cc @LiliC


s-urbaniak:
let me quickly check this on a vanilla 1.16 cluster before we merge.

PR #260, opened 2019-10-08
kube-prometheus/prometheus: Add local configuration

s-urbaniak:
This adds constructor'esque configuration options for prometheus assets.
They still reference global _config default values for backwards compatibility
but allow overriding values for new instances of prometheus assets.

/cc @LiliC @metalmatze @brancz


jsonnet/kube-prometheus/prometheus/prometheus.libsonnet
```diff
@@ -18,60 +18,67 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
...
-      service.new('prometheus-' + $._config.prometheus.name, { app: 'prometheus', prometheus: $._config.prometheus.name }, prometheusPort) +
+      service.new('prometheus-' + p.name, { app: 'prometheus', prometheus: p.name }, prometheusPort) +
       service.mixin.spec.withSessionAffinity('ClientIP') +
-      service.mixin.metadata.withNamespace($._config.namespace) +
-      service.mixin.metadata.withLabels({ prometheus: $._config.prometheus.name }),
-    [if $._config.prometheus.rules != null && $._config.prometheus.rules != {} then 'rules']:
+      service.mixin.metadata.withNamespace(p.namespace) +
+      service.mixin.metadata.withLabels({ prometheus: p.name }),
+
+    rules:
```
s-urbaniak: During a brainstom @metalmatze and me came to the conclusion that conditionally rendering that object adds unnecessary complexity. Instead we suggest to proactively hide this object in consumers if it is unneeded.

brancz:
The structure of this sgtm.

s-urbaniak:
Let me merge then, I can address additional comments as a follow up, if any.

BenoitKnecht:
@s-urbaniak I have the following config in my `default.jsonnet` file:

```jsonnet
local kp =
  (import 'kube-prometheus/kube-prometheus.libsonnet') +
  (import 'kube-prometheus/kube-prometheus-managed-cluster.libsonnet') +
  {
    _config+:: {
      prometheus+:: {
        name: 'custom',
      },
    },
  },
[...]
```

which used to work fine to set the Prometheus name, and played nice with [this part of kubernetes-grafana](https://github.com/brancz/kubernetes-grafana/blob/master/grafana/grafana.libsonnet#L15-L18).

When I upgraded `kube-prometheus`, I ended up with a second Prometheus server named `k8s`, which I now understand is caused by this PR.

Was it your intention to change this behavior? Should I change my configuration to work around this issue, or did you intend to keep retro-compatibility?

BenoitKnecht:
@s-urbaniak I've submitted #282 to address this particular issue. feel free to leave a comment there if you have any concerns.

s-urbaniak:
@BenoitKnecht indeed, you are correct, we should source the `name` and `replica` attributes from `_config` by default. Thanks a lot for submitting the patch!

PR #277, opened 2019-10-22
Add metrics_path label to kubelet servicemonitor endpoints

### LINKED ISSUE #276: Kubelet ServiceMonitor creates conflicting time series
**What happened?**
When we use the provided [ServiceMonitor for kubelet](https://github.com/coreos/kube-prometheus/blob/master/manifests/prometheus-serviceMonitorKubelet.yaml), each endpoint produces some identical metrics. In particular, this is the case for metrics that are generated for every scrape, such as `scrape_samples_post_metric_relabeling`. This is what happens when we graph this metric:

image

Based on the ServiceMonitor manifest, this is expected. The only difference between the endpoints is the metrics path, but this value is not added to the time series labels. However, despite this being expected, it's definitely a bug as it leads to the behavior shown in the image above.

Just apply the ServiceMonitor in any cluster that supports it.

-----------

At the end of the graph, you can see the lines smooth out. This was the result of adding this configuration into each of the endpoints:
```
relabelings:
- sourceLabels: [__metrics_path__]
  targetLabel: metrics_path
```

jutley:
Fixes #276


jutley:
Didn't realize that file is generated. I'll fix this later.

brancz:
Thanks! Otherwise I like this improvement a lot as we'll get distinct `up` metrics for the cadvisor and kubelet endpoints this way. Let us know if we can help with the generation part :)

jsonnet/kube-prometheus/kube-prometheus-insecure-kubelet.libsonnet
```diff
@@ -9,6 +9,12 @@
               scheme: 'http',
               interval: '30s',
               bearerTokenFile: '/var/run/secrets/kubernetes.io/serviceaccount/token',
+              relabelings: [
```
jutley: I'm not sure where this file is used, but I figured I'd update it for consistency's sake.

jutley:
@brancz I think this PR is good to go. Let me know if anything else needs adjusting.

brancz:
Thanks!

omerlh:
We also need to change the kublet/cadvisor selectors on the dashboards, using:
```
_config+:: {
      kubeletSelector: 'job="kubelet", metrics_path="/metrics"',
      cadvisorSelector: 'job="kubelet", metrics_path="/metrics/cadvisor"',
}
```

I can open a PR if it makes sense.

brancz:
Yes makes sense. Please go ahead! :)

omerlh:
here you go: coreos/kube-prometheus#360

PR #279, opened 2019-10-23
Use configured resources for kube-rbac-proxy in node-exporter daemonset

BenoitKnecht:
Since #132, `kube-rbac-proxy` resources can be configured in
`config.resources['kube-rbac-proxy']`, but the node-exporter daemonset
was still using hard-coded values.

This commit sets the request and limit resources to the configured
values for the `kube-rbac-proxy` container in the node-exporter pods.


brancz:
lgtm on green

brancz:
Thanks!

PR #282, opened 2019-10-24
prometheus: Let name and replicas be set in _config

BenoitKnecht:
Before #260, the Prometheus name and number of replicas could be
configured in `_config.prometheus.name` and
`_config.prometheus.replicas` respectively.

It isn't the case anymore, which means that configurations that did set
a custom name for Prometheus will get a second Prometheus instance
called `k8s` when they upgrade kube-prometheus (see https://github.com/coreos/kube-prometheus/pull/260#issuecomment-544981316).

This commit adds back the ability to configure both of these parameters
in `_config`.


brancz:
cc @s-urbaniak

s-urbaniak:
@BenoitKnecht great, catch, thank you! LGTM :+1:

PR #285, opened 2019-10-24
Makefile: Fix manifests target dependency

BenoitKnecht:
Since 1664600, manifests are built using `examples/kustomize.jsonnet`
instead of `example.jsonnet`.

This commit updates the dependencies in the `manifests` target to
reflect that change.


brancz:
Thanks!

PR #289, opened 2019-10-28
Fix memory saturation dashboard

### LINKED ISSUE #222: Memory Saturation (Swapped Pages) based on paging instead of swapping
**What happened?**

I looked at the Memory Saturation graph for a cluster/node that does not have any swap and noticed the graph indicated that swap was happening. `vmstat 1` on a node confirmed that there is no `si` or `so` activity.

I expected to see the graph only every showing 0 as no swap activity is possible on the nodes.

Create a node without swap. You may have to create a pod that actively uses mmapped files, which will create paging activity. Look at the graph and it will show the paging activity as swap.

**Environment**

    `quay.io/coreos/prometheus-operator:v0.32.0`
    

    `Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.2", GitCommit:"66049e3b21efe110454d67df4fa62b08ea79a19b", GitTreeState:"clean", BuildDate:"2019-05-16T18:56:40Z", GoVersion:"go1.12.5", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.2", GitCommit:"f6278300bebbb750328ac16ee6dd3aa7d3549568", GitTreeState:"clean", BuildDate:"2019-08-05T09:15:22Z", GoVersion:"go1.12.5", Compiler:"gc", Platform:"linux/amd64"}`
    

I know that this isn't defined in this project, but I got tired of trying to track down where it came from.

BenoitKnecht:
Upgrade to prometheus/node_exporter@d574b4b, which includes a better
metric for memory saturation.

Fixes #222

### DIFF
--- a/jsonnetfile.lock.json
+++ b/jsonnetfile.lock.json
@@ -97,7 +97,7 @@
                     "subdir": "docs/node-mixin"
                 }
             },
-            "version": "9f49fff79ef85fcebb69289622150e6d5346528b"
+            "version": "d574b4b41b72966a2ace4d6d81195081dfca8301"
         },
         {
             "name": "promgrafonnet",
--- a/manifests/grafana-dashboardDefinitions.yaml
+++ b/manifests/grafana-dashboardDefinitions.yaml
@@ -20553,7 +20553,7 @@ items:
                           "steppedLine": false,
                           "targets": [
                               {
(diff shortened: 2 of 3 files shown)


brancz:
Thanks!

PR #298, opened 2019-11-06
Disable controller and scheduler alerts in managed clusters

dctrwatson:
No reason to have `KubeControllerManagerDown` or `KubeSchedulerDown` alerts if we can't scrape them


brancz:
Thanks!

PR #307, opened 2019-11-18
Update README.md with apply clarification.

EricHorst:
Update the kubectl apply commands in the customizing section to match those the quickstart section. The customizing section did not account for the recently introduced setup/ subdirectory.


brancz:
Thanks!

PR #309, opened 2019-11-27
Group alert by namespace instead of job

benjaminhuo:
Some alert don't have job label which results in lots of ungrouped alerts


brancz:
How about just namespace? I always feel that grouping by alertname is not really helpful for myself. Grouping by namespace typically groups alerts in a somewhat meaningful way as a namespace is most often some kind of responsibility boundary.

benjaminhuo:
> How about just namespace? I always feel that grouping by alertname is not really helpful for myself. Grouping by namespace typically groups alerts in a somewhat meaningful way as a namespace is most often some kind of responsibility boundary.

OK，I'll make the change . Grouping by namespace only makes sense.
I added alertname previously, because I can see some alerts are triggered under the same alertname with difference labels such as pod, deployment etc.

benjaminhuo:
@brancz I've made the change to use namespace only for alert grouping

brancz:
It's an unrelated change, but this just needs a `make generate-in-docker` and commit the changes in a separate commit to fix CI. Sorry for the inconvenience.

benjaminhuo:
Finally, the CI is fixed

PR #330, opened 2019-12-16
 jsonnet,manifests: Drop all metrics which are deprecated in kubernetes

LiliC:
This PR drops all metrics that were deprecated in kubernetes versions 1.14 and 1.15. I went by the [chanelog](https://github.com/kubernetes/kubernetes/blob/0d58709016f3358f34a52c12087d707a64cc6cb0/CHANGELOG-1.17.md)/[s](https://github.com/kubernetes/kubernetes/blob/5e0f36705ef785a80955bd2b6412d9d227ff60da/CHANGELOG-1.14.md#removed-and-deprecated-metrics)

cc @brancz @s-urbaniak

### DIFF
--- a/jsonnet/kube-prometheus/dropping-deprecated-metrics-relabelings.libsonnet
+++ b/jsonnet/kube-prometheus/dropping-deprecated-metrics-relabelings.libsonnet
@@ -1,0 +1,50 @@
+[
+  // Drop all kubelet metrics which are deprecated in kubernetes.
+  {
+    sourceLabels: ['__name__'],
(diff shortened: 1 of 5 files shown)


brancz:
This PR only drops metrics from the kubelet but the dropped metrics indicate they also come from other components. Am I missing something?

LiliC:
@brancz fixed, thanks! PTAL

brancz:
Nice :+1: I would expect us to reduce metrics quite a lot here as some of the highest cardinality ones are in there. 

lgtm

PR #332, opened 2019-12-18
jsonnet/kube-prometheus/jsonnetfile.json: Pin prometheus-operator version to master instead

LiliC:
We should not pin against a release in master.

cc @s-urbaniak


LiliC:
Forgot to update some deps, will push changes. 

/hold

s-urbaniak:
agreed, let's unpin here :+1:

PR #336, opened 2019-12-19
jsonnet/kube-prometheus: Adjust dropped deprecated metrics names

LiliC:
The names were not complete in the kubernetes CHANGELOG. I manually checked these against a running cluster to make sure this is the full name of the metric. In the CHANGELOG it was lacking the prefix. 🤷‍♀ 

cc @brancz @s-urbaniak

### DIFF
--- a/jsonnet/kube-prometheus/dropping-deprecated-metrics-relabelings.libsonnet
+++ b/jsonnet/kube-prometheus/dropping-deprecated-metrics-relabelings.libsonnet
@@ -2,7 +2,7 @@
   // Drop all kubelet metrics which are deprecated in kubernetes.
   {
     sourceLabels: ['__name__'],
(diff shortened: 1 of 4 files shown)


brancz:
Not completely fault proof but we can extract all metadata about metrics in a running cluster and ensure that there are no deprecated ones. Would make a nice smoke test here I think just to ensure that we never use deprecated metrics.

LiliC:
One second, found a problem.
/hold

LiliC:
@brancz do you want me to do that in this PR?

brancz:
I’m happy with a manual verification for this PR but would be nice as a follow up.

LiliC:
@brancz will open an issue as a follow up to do that.

s-urbaniak:
lgtm ... fwiw :-)

PR #346, opened 2019-12-23
fix coredns monitoring on EKS

omerlh:
On EKS, CoreDNS does not have service endpoint with the name `metrics`. Using `targetPort` instead solved the issue and allow Prometheus to discover core DNS.


brancz:
Thanks!

PR #361, opened 2020-01-06
Avoid base64-encoding of default grafana-datasources secret

erwinvaneyk:
The default `grafana-datasources ` config is currently stored in base64-encoded form. This makes it hard to see what the configuration is and to update it if needed.


brancz:
This file is auto-generated from jsonnet, so the change would need to be made there, and in fact the source jsonnet is not even in this repo but in https://github.com/brancz/kubernetes-grafana. That said, I think this is a great suggestion, and I'd be happy to see this change in the kubernetes-grafana repo.

metalmatze:
I think the reason this was created as a secret is the fact that you might have some passwords / tokens in that configuration as part of basic auth, for example. Not sure a ConfigMap would be sufficient.

brancz:
I don't think the proposal was to use a configmap, but rather just the `stringData` field of the `Secret`.

erwinvaneyk:
The proposal is indeed to use `stringData` in the secret for the default setup to make it more readable/understandable. Thanks for the pointer, @brancz - I will open a PR in https://github.com/brancz/kubernetes-grafana :+1:

metalmatze:
Oh ok! That sounds like a good improvement :blush:

ArthurSens:
Looks like this one has already been fixed by https://github.com/prometheus-operator/kube-prometheus/pull/1364 🙂, so I'm closing this one

PR #367, opened 2020-01-09
tests/e2e/travis-e2e.sh: Switch to 1.17 k8s cluster

LiliC:
🤞 :)


brancz:
Nice! Actually kind of incredible that there were no other metrics to deprecate ...

LiliC:
@brancz it's because I was testing against a 1.16 cluster locally and those were already included.

PR #374, opened 2020-01-13
Transform custom-metrics into an addon

johanneswuerbach:
Instead of keeping the custom-metrics-api separate and experimental https://github.com/coreos/kube-prometheus/tree/master/experimental/custom-metrics-api, move it into jsonnet and use the same already deployed [k8s-prometheus-adapter](https://github.com/DirectXMan12/k8s-prometheus-adapter/). The adapter supports custom-metrics (and also external-metrics) for awhile now and we are using it in production without issues.

What do you think?

I'm happy to re-generate the manifests if this looks reasonable.


LiliC:
Thanks for your PR. 
cc @s-urbaniak as you might know more about this.

brancz:
I don't think we're ready to do this yet. None of the scalability limitations around the custom/external metrics adapter part of it have been solved, that were the reason for us not including it by default.

s-urbaniak:
yes, agreed. I share the same assertion that prometheus-adapter is not production ready yet regarding custom and external metrics.

johanneswuerbach:
> I share the same assertion that prometheus-adapter is not production ready yet regarding custom and external metrics.

Do you have further details on what is currently missing / a list of open items? Are they affecting both custom/external metrics or just one of them?

metalmatze:
Some plans for the future were mentioned in our talk at KubeCon: 
https://youtu.be/cIoOAbzhR7k?t=1396

But hopefully, there are some issues for people to start with.

brancz:
I believe @s-urbaniak started writing a design doc for these improvements. Can we maybe publish it in the state that we have it even if incomplete?

s-urbaniak:
The design started it's live here https://github.com/s-urbaniak/prometheus-adapter/blob/master/design.md, along side with an initial implementation (that actually also booted up).

The configmap proposal in that document for resource metrics is actually something I envisioned (aka a nearly zero-config expression of a sample query without templating)

The CRD design for custom and external metrics is TBD, but would follow the same idea above.

johanneswuerbach:
How would you feel about making this an opt-in patch instead so it easier for users to keep up-to-date and use it together with kube-prometheus jsonnet?

brancz:
I think an opt-in patch is ok as people would make the choice explicitly.

jsonnet/kube-prometheus/kube-prometheus-custom-metrics.libsonnet
```diff
@@ -0,0 +1,141 @@
...
+
+// Custom metrics API allows the HPA v2 to scale based on arbirary metrics.
+// For more details on usage visit https://github.com/DirectXMan12/k8s-prometheus-adapter#quick-links
+
+{
+  _config+:: {
+    prometheusAdapter+:: {
+      // Rules for custom-metrics
+      config: |||
+        rules:
```
brancz, line 11: to avoid duplication, how about we make this config not an opaque string, but rather structured, so we can just merge the additonal configuration in, instead of having to re-specify it, which is prone to error when we fix things in one place but not the other

johanneswuerbach: This would also mean adjusting https://github.com/coreos/kube-prometheus/blob/master/jsonnet/kube-prometheus/prometheus-adapter/prometheus-adapter.libsonnet#L20, correct?

Happy to give it a try :-)

brancz: correct

johanneswuerbach: Made the change of the config in https://github.com/coreos/kube-prometheus/pull/485, will update this PR once this looks good :-)

brancz: Merged that :)

johanneswuerbach: @brancz updated :-)

brancz:
I’m happy with this especially as it’s an opt in. Giving @lilic and @s-urbaniak also a chance to review.

johanneswuerbach:
Friendly ping to @lilic and @s-urbaniak :-)

No idea why travis says this is pending as clicking on details actually says it passed 🤔

lilic:
> No idea why travis says this is pending as clicking on details actually says it passed 🤔

We have been seeing this with travis lately, it does not report the status back.

jsonnet/kube-prometheus/kube-prometheus-custom-metrics.libsonnet
```diff
@@ -0,0 +1,179 @@
+local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
```
lilic, line 1: Do we not want to also generate the files coming from this as an example users can try out, same way as we had it before? @brancz ?

johanneswuerbach: The feedback was that custom-metrics aren't fully ready yet and so this addon should only be opt-in vs. installed by default https://github.com/coreos/kube-prometheus/pull/374#issuecomment-573692256

lilic: Right now we are removing the example experimental directory, all I would want is the files from this libsonnet to be generated by default same in the same way as they were before in the experimental directory. I agree it should be opt-in, but this just makes it harder for folks to try out.

johanneswuerbach: Make sense, I added a new commit to generate the experimental manifest from jsonnet.

examples/custom-metrics-api.jsonnet
```diff
@@ -0,0 +1,11 @@
+local kp =
+  (import 'kube-prometheus/kube-prometheus.libsonnet') +
+  (import 'kube-prometheus/kube-prometheus-custom-metrics.libsonnet') +
+  {
+    _config+:: {
+      namespace: 'monitoring',
+    },
+  };
+
+{ ['setup/0namespace-' + name]: kp.kubePrometheus[name] for name in std.objectFields(kp.kubePrometheus) } +
```
johanneswuerbach, line 10: Technically you would also need a prometheus, the operator etc. Not sure if all of this should also be generated (it wasn't before)

jsonnet/kube-prometheus/kube-prometheus-custom-metrics.libsonnet
brancz: We don't generate any of the other additive things like kops, kubespray, eks, aks, gke, ... I don't think we should start doing this for any other add-ons as it sets the wrong precedence in my opinion. We should focus on driving people towards generating in my opinion.

johanneswuerbach: Happy to remove the 2nd commit again, if that is the agreement :-)

lilic: Sorry about that! 🤷 up to frederic I guess, we discussed this OOB and feel free to remove the commit! 

But @brancz needs to open an issue about testing the manifests we ship or at least producing the manifests as part of CI runs so we can inspect any problems.

johanneswuerbach: @lilic no worries, removed :-)

brancz:
Thanks a ton for bearing with us! :) :tada:

PR #390, opened 2020-01-24
KubeNodeUnreachable fires after 2 minutes now

amir-hadi:
I propose this change because we are using the Kubernetes Cluster Autoscaler and whenever the autoscaler terminates nodes, for a brief amount of time the alert `KubeNodeUnreachable` is firing. I think this could solve it, but any feedback is appreciated.

Does anyone else have this problem with this alert? Maybe something is wrong in our EKS 1.14 cluster.

What do you think?

### DIFF
--- a/manifests/prometheus-rules.yaml
+++ b/manifests/prometheus-rules.yaml
@@ -1043,6 +1043,7 @@ spec:
         runbook_url: https://github.com/kubernetes-monitoring/kubernetes-mixin/tree/master/runbook.md#alert-name-kubenodeunreachable
       expr: |
         kube_node_spec_taint{job="kube-state-metrics",key="node.kubernetes.io/unreachable",effect="NoSchedule"} == 1
+      for: 2m
       labels:
         severity: warning
     - alert: KubeletTooManyPods


brancz:
@smarterclayton you intentionally added the alert as is without a for clause. And if I recall correctly you said something also the lines of that if this fires unexpectedly then it’s more likely to be a bug elsewhere. Can you refresh my memory and let us know where to open what bug?

amir-hadi:
The tests are failing. I don't completely understand how they work and where the data comes from, but if somebody tells me I would fix it :-).

amir-hadi:
@brancz We can directly correlate the alert with cluster nodes being not needed anymore and they get shutdown from the autoscaler. In that moment this alert fires.

Could this be a bug in the autoscaler? How it shuts down nodes?

brancz:
The alert is not actually defined in this repo it's generated into the file that you changed and CI re-generates and overrides your changes. The alert is pulled in via this dependency: https://github.com/kubernetes-monitoring/kubernetes-mixin

amir-hadi:
@smarterclayton If you could give me some more insights if this is intended? For now we disabled the alert, but I'd like to know if it actually should not occur. If you say it is intended, we can close this. If you say, we can add the `for` statement, I'd create a PR in the `kubernetes-mixin` project.

brancz:
@amir-hadi could you reply to this thread regarding the questions that @smarterclayton asked?  https://github.com/kubernetes-monitoring/kubernetes-mixin/issues/300

amir-hadi:
I did, I am closing this now as this is not the right place anyway. Thanks for the help @brancz!

PR #391, opened 2020-01-24
*: Add some simple default inhibition rules

brancz:
I had these laying in my notes for a while and just wanted to get rid of it in my head :) 

For readability I also converted this from being base64 encoded data to `stringData`.

@s-urbaniak @LiliC @simonpasquier @paulfantom @pgier


benjaminhuo:
@brancz Just wondering if inhibiting by alertname while grouping by namespace will be a problem?

brancz:
@benjaminhuo it should, but you're right, it's probably worth adding `namespace` to the `equal` array, just to make sure an alert in one namespace isn't inhibiting an alert in another namespace. Do you want to create a PR to add it?

benjaminhuo:
Sure

PR #400, opened 2020-01-30
Include service in targetdown

### LINKED ISSUE #394: TargetDown alert isn't specific enough when job includes multiple services
**What is missing?**
When a job encompasses multiple services, the [target down message](https://github.com/coreos/kube-prometheus/blob/master/jsonnet/kube-prometheus/alerts/general.libsonnet#L10) isn't specific enough to identify which service has down targets.

**Why do we need it?**
Allowing jobs with multiple targets / services seems reasonable. The alert already groups `BY (job, namespace, service)`.

**Environment**
* kube-prometheus version:  master

I would propose that the message be changed to read `the {{ $labels.job }}/{{ $labels.service }} targets`. I am more than happy to create that PR if that seems reasonable.

JTarasovic:
This should fix #394

### DIFF
--- a/jsonnet/kube-prometheus/alerts/general.libsonnet
+++ b/jsonnet/kube-prometheus/alerts/general.libsonnet
@@ -7,7 +7,7 @@
           {
             alert: 'TargetDown',
             annotations: {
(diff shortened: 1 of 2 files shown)


brancz:
lgtm :+1: thanks!

PR #408, opened 2020-02-07
Remove preserveUnknownField CRD setting

pgier:
This allows the CRDs to be generated to maintain compatibility with
Kubernetes 1.14 and earlier versions.

Helps to mitigate https://github.com/coreos/prometheus-operator/issues/2924

### DIFF
--- a/jsonnet/kube-prometheus/kube-prometheus.libsonnet
+++ b/jsonnet/kube-prometheus/kube-prometheus.libsonnet
@@ -17,6 +17,34 @@ local configMapList = k3.core.v1.configMapList;
   kubePrometheus+:: {
     namespace: k.core.v1.namespace.new($._config.namespace),
   },
+  prometheusOperator+::
+  {
+    '0alertmanagerCustomResourceDefinition'+: {
+      spec: std.mergePatch(super.spec, {
+        preserveUnknownFields: null,
+      }),
+    },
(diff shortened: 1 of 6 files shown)


pgier:
I originally tried to make this configurable, but couldn't find a nice way to do it in jsonnet that would fit all the use-cases.
I think it might be better anyway to just always remove preserveUnknownFields from the CRDs.  The only drawback is that `kubectl explain` won't work on newer versions of kube (1.15+), but I think that's not as important as maintaining compatibility with earlier kube versions for now.

s-urbaniak:
lgtm

omerlh:
I'm having a similar issue with `x-kubernetes-int-or-string: true`, which also is not available on Kubernetes 1.14. it will be a bit harder to remove it... In this case, the apply succeed because Kubernetes ignore this field - but it is problematic when using GitOps tool like arogcd/flux - as the field exist on git, but not on cluster.

brancz:
I don’t think we want this on master do we? I think master should always represent the compatibility with the most up to date versions with the best known practices.

pgier:
@omerlh Would you mind opening a new issue for that with the version(s) of kube you are using and the error messages you are seeing?

omerlh:
Sorry for the delate, see #506

PR #432, opened 2020-03-03
Adjust the threshold of the NodeFilesystemSpaceFillingUp alert from the node-exporter mixin

iuriaranda:
Reduce threshold of the node-exporter alert to 15% space available, instead of 20% (default).

As per https://github.com/coreos/kube-prometheus/issues/294


brancz:
Thanks! lgtm :+1:

PR #436, opened 2020-03-04
jsonnet/kube-prometheus/kube-state-metrics: Remove probes

lilic:
cc @coreos/team-monitoring @dgrisonnet 

Still want to test for longer, but ready for review. Is there another way other then doing null in jsonnet to remove it? @brancz @metalmatze

### DIFF
--- a/jsonnet/kube-prometheus/kube-state-metrics/kube-state-metrics.libsonnet
+++ b/jsonnet/kube-prometheus/kube-state-metrics/kube-state-metrics.libsonnet
@@ -33,7 +33,9 @@
                             template+: {
                               spec+: {
                                 containers: std.map(function(c) c {
-                                  ports: null,
+                                  ports:: null,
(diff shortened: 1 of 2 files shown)


jsonnet/kube-prometheus/kube-state-metrics/kube-state-metrics.libsonnet
```diff
@@ -34,6 +34,8 @@
                               spec+: {
                                 containers: std.map(function(c) c {
                                   ports: null,
+                                  livenessProbe: null,
+                                  readinessProbe: null,
```
metalmatze:
```suggestion
                                  livenessProbe:: null,
                                  readinessProbe:: null,
```
Would actually hide them.

lilic:
Ready to be merged, no restarted containers locally and no error logs in the proxy containers.

PR #441, opened 2020-03-06
fix invalid Usage section reference

jadia:
Replace Usage section reference to Customizing Kube-Prometheus section.

Usage section was removed from the README.md file in [this](https://github.com/coreos/kube-prometheus/commit/e8e0b639f7098116f558422d40a0433d851d6e20) commit. But the reference to the section was not modified.


brancz:
Thanks!

PR #447, opened 2020-03-11
Update prometheus-rules.yaml

mcanevet:
Ignore KubeletTooManyPods on fargate nodes as it's its normal behavior.


lilic:
This comes from the mixin repo https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/12773f7181057163a3761913bb629d7ce9e13bca/alerts/kubelet.libsonnet#L39-L46 Please open PR there.

PR #449, opened 2020-03-12
Ignore KubeletTooManyPods on Fargate nodes

mcanevet:
Ignore KubeletTooManyPods on fargate nodes as it's its normal behavior.
This replaces #447.


mcanevet:
@lilic superseded by https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/377

PR #450, opened 2020-03-12
adding security context to kube-rbac-proxy

latchmihay:
- security contexts were dropped during the latest update/rewrite of the `kube-rbac-proxy`


brancz:
Thanks for the PR! Why not handle this via PodSecurityPolicies?

latchmihay:
> Thanks for the PR! Why not handle this via PodSecurityPolicies?

I am not sure how I would handle that trough PSP? In fact I have PSPs that prevent it to run as `root` thats why I am implementing the security context in this PR. It was there before and was dropped during the last rewrite

brancz:
Ah sorry, I tend to mistake (OpenShift) SecurityContextConstraints (which enforce/override secure values) and PodSecurityPolicies. This lgtm on green CI (it needs a re-generate with `make clean generate-in-docker`).

PR #454, opened 2020-03-17
Add k8s-app label for the kube-state-metrics ServiceMonitor

rajatvig:


jsonnet/kube-prometheus/kube-state-metrics/kube-state-metrics.libsonnet
```diff
@@ -52,6 +52,7 @@
                               labels: {
                                 'app.kubernetes.io/name': 'kube-state-metrics',
                                 'app.kubernetes.io/version': ksm.version,
+                                'k8s-app': 'kube-state-metrics',
```
lilic: What is the need for adding this, we already have `app.kubernetes.io/name`which is the official recommended label https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels

rajatvig: The `k8s-app` label is applied to all other service monitors in this repository. This just makes it consistent with everything else.

lilic: Okay thanks, makes sense
cc @brancz any objections?

brancz: It's not a standardized label, I don't think we should continue using it.

rajatvig: @brancz Our target discovery has ended up relying on that label but I see the point you are making.
Would it make sense then to remove all the other instances and allow it to be overridden?
Can rework the PR to that effect.

brancz: I think we can work on each component one by one to move to the recommended app labels yes. I would prefer if we can do that one by one for each component one PR for each.

rajatvig: 👍

rajatvig:
@brancz @lilic Here is the first two PRs
1. https://github.com/coreos/kube-prometheus/pull/458
2. https://github.com/coreos/kube-prometheus/pull/459

PR #462, opened 2020-03-20
Fix kube-state-metrics namespace override

### LINKED ISSUE #460: kube-state-metrics namespace is hardcoded to "monitoring"
This is a regression introduced in #404.

From `kube-state-metrics/kube-state-metrics.libsonnet`:

```jsonnet
  kubeStateMetrics+:: (import 'kube-state-metrics/kube-state-metrics.libsonnet') +
                      {
                        local ksm = self,
                        name:: 'kube-state-metrics',
                        namespace:: 'monitoring',
```

The namespace here should be `$_config.namespace` (which defaults to `'monitoring'`, but can be changed).

bgagnon:
Use `$._config.namespace` instead of hard-coding `'monitoring'`.

Fixes #460


brancz:
Thanks!

PR #463, opened 2020-03-21
Support standard labels for nodeexporter

rajatvig:


manifests/node-exporter-daemonset.yaml
```diff
@@ -3,6 +3,8 @@ kind: DaemonSet
 metadata:
   labels:
     app: node-exporter
```
brancz: this `app` label shouldn't be here anymore now I think

rajatvig: Nice catch. Fixed it and resolved conflicts.

rajatvig:
Had to update the manifests other than those for node exporter as they seem out of date.

brancz:
Lgtm but giving @lilic the last call.

lilic:
Thanks!

PR #471, opened 2020-03-24
Append the ports for the Operator vs replacing

rajatvig:
The last change to the Operator service overwrites the ports from the Service vs appending them.


brancz:
Why do you want to access the Prometheus Operator without TLS?

@paulfantom my understanding is that this was intentional, right?

PR #472, opened 2020-03-26
Secure prometheus-k8s metrics

dgrisonnet:
Related to #468


dgrisonnet:
@paulfantom I added a prom-label-proxy behind a rbac-proxy whose dropping the secure connection with Prometheus since it is not needed anymore. However, because of that, I can't have the e2e tests to pass, do you have any ideas on how I could tweak that?

PR #478, opened 2020-04-02
Remove version label from node-exporter selectors

Capitrium:
The `spec.selector` field is immutable on DaemonSets, so #463 ended up breaking the upgrade process (i.e. `kubectl apply -f manifests/`). This PR removes the version label from the selector, which fixes the issues with applying changes.

### DIFF
--- a/jsonnet/kube-prometheus/node-exporter/node-exporter.libsonnet
+++ b/jsonnet/kube-prometheus/node-exporter/node-exporter.libsonnet
@@ -20,6 +20,11 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
         'app.kubernetes.io/name': 'node-exporter',
         'app.kubernetes.io/version': $._config.versions.nodeExporter,
       },
+      selectorLabels: {
+        [labelName]: $._config.nodeExporter.labels[labelName]
+        for labelName in std.objectFields($._config.nodeExporter.labels)
+        if !std.setMember(labelName, ['app.kubernetes.io/version'])
+      },
     },
   },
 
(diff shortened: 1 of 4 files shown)


jsonnet/kube-prometheus/node-exporter/node-exporter.libsonnet
```diff
@@ -69,6 +69,11 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
       local containerEnv = container.envType;
 
       local podLabels = $._config.nodeExporter.labels;
+      local selectorLabels = {
```
brancz, line 72: let's move this to the config section and make use of this in the service selector as well

Capitrium: The service selector is already picking up these change since it just refers to the daemonset's selector (`$.nodeExporter.daemonset.spec.selector.matchLabels`) but I can definitely make that more explicit. So you'd prefer if both the daemonset and the service used something like `$._config.nodeExporter.selectorLabels` instead?

brancz: Yeah and have the ServiceMonitor use the same.

brancz:
This needs a “make generate-in-docker”.

brancz:
Nice! lgtm :+1: 

Giving @lilic a chance to review this as well though :)

johanneswuerbach:
Any chance this could be merged? This also breaks our update.

lilic:
lgtm, sorry for missing this. :) Thanks for the ping!

PR #485, opened 2020-04-07
Make prometheus-adapter config a real object

johanneswuerbach:
This allows proper merging adapter configs instead of working with strings.

So modifying the config becomes for addons like https://github.com/coreos/kube-prometheus/pull/374


brancz:
No clue why Travis CI is not reporting status, but it's green.

lgtm :+1: thanks! :tada:

jsonnet/kube-prometheus/prometheus-adapter/prometheus-adapter.libsonnet
```diff
@@ -16,34 +16,47 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
...
+                  resource: 'namespace'
+                },
+                pod: {
+                  resource: 'pod'
+                },
+              },
+            },
+            containerLabel: 'container'
+          },
+          window: 5,
```
yeya24, line 57: This window configuration should be `5m`. I don't know why this happened, is there a CI for it? I will open a pr to fix it

PR #488, opened 2020-04-07
prometheus-adapter: Fix rules window

johanneswuerbach:
This was regressed in https://github.com/coreos/kube-prometheus/pull/485


brancz:
Thanks!

PR #496, opened 2020-04-16
Bump dependencies 

lilic:
This is to bump all the dependencies before the release is cut. 

cc @coreos/team-monitoring @dgrisonnet

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ dist: xenial
 language: go
 
 go:
-- "1.12.x"
+- "1.13.x"
 go_import_path: github.com/coreos/kube-prometheus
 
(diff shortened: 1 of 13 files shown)


lilic:
Ready for review @coreos/team-monitoring @dgrisonnet  PTAL, had to bump the go version as well as that was the only way tests passed for, as i am using 1.13.

brancz:
I expected to see a pin to kubernetes-mixin `release-0.4`. Any reason why it's not here?

lilic:
I guess I can do that in the same PR, was thinking of doing a bump and another of for cutting the release itself.

brancz:
That works for me as well.

PR #499, opened 2020-04-17
Pin dependencies for the 0.5 release branch

lilic:
cc @coreos/team-monitoring

### DIFF
--- a/jsonnet/kube-prometheus/jsonnetfile.json
+++ b/jsonnet/kube-prometheus/jsonnetfile.json
@@ -8,7 +8,7 @@
           "subdir": "grafana"
         }
       },
-      "version": "master"
+      "version": "release-0.1"
     },
     {
       "source": {
@@ -17,7 +17,7 @@
           "subdir": "Documentation/etcd-mixin"
         }
       },
-      "version": "master"
+      "version": "release-3.4"
     },
     {
       "source": {
@@ -45,7 +45,7 @@
           "subdir": ""
         }
       },
-      "version": "master"
+      "version": "release-0.4"
     },
     {
       "source": {
@@ -54,7 +54,7 @@
           "subdir": "jsonnet/kube-state-metrics"
         }
       },
-      "version": "master"
+      "version": "release-1.9"
     },
     {
       "source": {
@@ -63,7 +63,7 @@
           "subdir": "jsonnet/kube-state-metrics-mixin"
         }
       },
-      "version": "master"
+      "version": "release-1.9"
     },
     {
       "source": {
@@ -72,7 +72,7 @@
           "subdir": "docs/node-mixin"
         }
       },
-      "version": "master"
+      "version": "v0.18.1"
     },
     {
       "source": {


README.md
```diff
@@ -89,11 +89,12 @@ $ minikube addons disable metrics-server
...
 ### Kubernetes compatibility matrix
 
-| kube-prometheus stack | Kubernetes 1.14 | Kubernetes 1.15 | Kubernetes 1.16 | Kubernetes 1.17 |
-|-----------------------|-----------------|-----------------|-----------------|-----------------|
+| kube-prometheus stack | Kubernetes 1.14 | Kubernetes 1.15 | Kubernetes 1.16 | Kubernetes 1.17 | Kubernetes 1.18 |
+|-----------------------|-----------------|-----------------|-----------------|-----------------|-----------------|
 | `release-0.3`         | ✔              | ✔              | ✔              | ✔              |
 | `release-0.4`         | ✗              | ✗              | ✔              | ✔              |
-| `HEAD`                | ✗              | ✗              | ✗              | ✔              |
+| `release-0.5`         | ✗              | ✗              | ✗              | ✔              | ✔
```
brancz, line 96: Any particular reason for the 1.16 (+downward) incompatibility? Mainly curious, I'm also happy with the answer being "let's just not care for more releases, and encourage people to upgrade".

lilic: We do not test that in this branch, that is the only reason, we can't guarantee it, in the other branches we run our e2e tests against those k8s versions so we can say it passes.

At some point we have to decide how many versions we want to support.

lilic:
Want to bump k8s version we test against in master first to 1.18 and bring that commit here.

brancz:
lgtm

dgrisonnet:
You will also need to run the tests on `1.17` like we did here: #476

README.md
brancz: just wanted to make sure we're aligned, totally ok with this

lilic:
Yep, waiting for master PR to merge first to bring that commit in only figured out we do not test against 1.18 when I was editing the README, hence the WIP :)

lilic:
Decided to merge this first, to unblock CMO stuff, as 1.18 in master is taking a bit longer. @dgrisonnet @brancz this just pins deps PTAL, thanks!

jsonnet/kube-prometheus/jsonnetfile.json
```diff
@@ -72,7 +72,7 @@
           "subdir": "docs/node-mixin"
         }
       },
-      "version": "master"
+      "version": "v0.18.1"
```
dgrisonnet, line 75: Is there a reason why we are using a tag instead of the branch `release-0.18` here?

lilic: actually noticed that this does not work at all, but somehow our CI did not catch this. Before with make generate-in-docker we ran jb update/install now we skip this step so when running make generate it did not catch this, I am already fixing this.

dgrisonnet: I looked into it and it appears that we never ran `jb update` in the CI: https://travis-ci.org/github/coreos/kube-prometheus/jobs/674766820#L261.
It is not linked to moving away from `generate-in-docker` but to the lack of `jb update`

lilic: yes, true! My mistake, this should fix it https://github.com/coreos/kube-prometheus/pull/502

PR #502, opened 2020-04-17
Fix broken pinning

lilic:
Well someone (me 🤚) forgot to run jb update :)  This fixes that...

cc @brancz @dgrisonnet


brancz:
lgtm on green

PR #512, opened 2020-04-24
Add ability to insert args to node-exporter container

Briggybros:
Useful if you'd like to enable a collector from https://github.com/prometheus/node_exporter#disabled-by-default


brancz:
I think this should be done with plain jsonnet as a patch in downstream usage.

PR #481, opened 2020-04-06
Allow to configure EKS available IPs alert

omerlh:
Make EKS available IP configureable

PR #513, opened 2020-04-27
Allow to configure EKS available IPs alert

### LINKED ISSUE #506: x-kubernetes-int-or-string CRD field is not support on k8s 1.14
**What happened?**
The CRD definition generated by the mixin library add this field `x-kubernetes-int-or-string` which is supported only on k8s 1.15 or higher. Worth mentioning that until recently, 1.14 was the latest version supported by EKS. 

The field is ignored by k8s, but when using tools like Argo for deployment, you'll see a constant difference between the desired state and the actual state.

Option to remove this field.

Deploy the CRD definitions on a cluster with version 1.14, notice that the field is removed.

omerlh:
Merge #481 into release-0.3 and close #506


dgrisonnet:
lgtm

PR #519, opened 2020-05-05
Revert "Remove field preserveUnknownFields from CRDs"

pgier:
The preserveUnknownFields flag was removed previously to maintain support for kube v1.14.  With the move to CRD v1 (from v1beta1) we will no longer be able to support kube 1.14 and earlier, so I think it makes sense now to revert the previous change.

See https://github.com/coreos/prometheus-operator/issues/3178


pgier:
cc @brancz @dgrisonnet

dgrisonnet:
lgtm

brancz:
Looks good at first sight but would like another review @lilic.

PR #523, opened 2020-05-08
WIP: F/merge manifsets from helm chart

### LINKED ISSUE #522: WIP: Structure of manifests vs helm chart
- ./manifests vs ./jsonnet/kube-prometheus differs greatly
- our opinion is that plain k8s manifests should be the foundation for everything, as it's not plausible to learn jsonnet on top of learning k8s/ops/dev/etc...
- therefore we focus on making a kick-ass manifests folder to use

haf-afa:
Fixes #522


brancz:
The manifests are generated by jsonnet, and that should stay that way, other than that these look like awesome improvements! I would suggest to break this up into a series of smaller PRs that way things are easier to review and thus land quicker :) my suggestion would be to start with the restructuring as everything else more or less can build on top of that.

PR #535, opened 2020-05-16
Add new variable {node selector}

Asgoret:
**Feature**
Add new variable node selector to applications for controlling nodes for deploying. 

**For what**
For example, you have several zones like DMZ & internal. You don't want to deploy a monitoring system in the DMZ zone, so you need to change the default node selector.

**TODO**

- [ ] Prometheus
- [ ] Prometheus Adapter
- [ ] Alertmanager
- [ ] kube metrics server

Note: Grafana in another repository. For her will be another PR.

**UPD#1**: Broke deployment in OKD. Will be re-write.

### DIFF
--- a/jsonnet/kube-prometheus/alertmanager/alertmanager.libsonnet
+++ b/jsonnet/kube-prometheus/alertmanager/alertmanager.libsonnet
@@ -14,6 +14,7 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
 
     alertmanager+:: {
       name: 'main',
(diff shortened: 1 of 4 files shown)


brancz:
This is pretty simple to do as a patch in downstream usage no? I'm not sure we should have this configurable like this.

Asgoret:
@brancz well...it's simple, but strange that it is not done yet. Anyway, there is a lot of changes for deploying in OKD (Openshift 3.x), so I simply write documentation for it with all steps (if you needed here :) )

brancz:
> so I simply write documentation for it with all steps (if you needed here :) )

Docs would be awesome!

PR #536, opened 2020-05-18
Release 0.4 prom oper v0.35.1

pgier:
I believe the 0.4 release branch should have been pinned to prometheus operator v0.35 instead of the previous 0.34.


lilic:
cc @paulfantom you did the release for openshift 4.4 https://github.com/openshift/cluster-monitoring-operator/pull/624 (https://github.com/openshift/prometheus-operator/pull/70#issuecomment-630457278)

pgier:
Looks like the CRD changes here cause a problem in the API server in kube 1.16.1 and maybe some other early 1.16.x versions.  This looks to be similar to #411.  Applying any of the prometheus CRDs followed by a `kubectl get <crdtype>` causes the API server to start throwing nil pointer exceptions.  The same problem doesn't occur with the latest patch release of kube 1.16.9.

@lilic what are our requirements for the 1.16 support?  Can we just upgrade to kube 1.16.9?

lilic:
> what are our requirements for the 1.16 support? Can we just upgrade to kube 1.16.9?

You need to check which version we ship with 4.4.

pgier:
@lilic [openshift 4.4](https://docs.openshift.com/container-platform/4.4/release_notes/ocp-4-4-release-notes.html#ocp-4-4-about-this-release) is based on kube 1.17 so it shouldn't be an issue for downstream openshift compatibility.

I guess we could add a note to the compatibility matrix for this branch about requiring at least a certain minimum patch version of 1.16?

pgier:
@lilic Looks like this is the bug that causes the failure (https://github.com/kubernetes/kubernetes/pull/83789) and it was fixed in v1.16.2.  I'm thinking we should just update the compatibility matrix to require at least v1.16.2.

PR #549, opened 2020-05-26
Pin kubernetes-mixin version in release-0.3

dgrisonnet:
kubernetes-mixin master branch is only compatible with Kubernetes 1.18+, thus breaking this branch Kubernetes compatibility.

I pinned it according to what we have downstream: https://github.com/openshift/cluster-monitoring-operator/blob/release-4.3/jsonnet/jsonnetfile.json#L23-L29

### DIFF
--- a/jsonnet/kube-prometheus/jsonnetfile.json
+++ b/jsonnet/kube-prometheus/jsonnetfile.json
@@ -18,7 +18,7 @@
                     "subdir": ""
                 }
             },
-            "version": "master"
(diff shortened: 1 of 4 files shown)


brancz:
Lgtm but would like another pair of eyes

elisiano:
I thought `release-0.3` of kubernetes-mixin would be compatible?

elisiano:
never mind me, I thought this was the "other" PR

PR #570, opened 2020-06-11
Update endpoints

stafot:
Update endpoints on `prometheus-adapter` and `alertmanager` manifests to avoid `NXDOMAIN`s. 
I am fairly new to `jsonnet` so if there is any better way to do it feel free to correct me. 
I am sharing the following images to explain the reasoning behind this pull request. 
In the following image is shown the default setup of `alertmanager` that causes a noticeable amount of `NXDOMAIN` on dns lookups. 

alertmanager_endpoint

After we applied the proposed `alertmanager` changes on our setup these `NXDOMAIN`s were disappeared which improved our cluster performance by avoiding 3 `NXDOMAIN` per second. The same reasoning applies to the proposed change on `prometheus-adapter`.

fixed_alertmanager_endpoints

### DIFF
--- a/docs/developing-prometheus-rules-and-grafana-dashboards.md
+++ b/docs/developing-prometheus-rules-and-grafana-dashboards.md
@@ -252,30 +252,32 @@ local kp = (import 'kube-prometheus/kube-prometheus.libsonnet') + {
   _config+:: {
     namespace: 'monitoring',
   },
-  grafanaDashboards+:: {
-    'my-dashboard.json':
-      dashboard.new('My Dashboard')
-      .addTemplate(
-        {
-          current: {
-            text: 'Prometheus',
-            value: 'Prometheus',
+  grafana+:: {
+    dashboards+:: {
+      'my-dashboard.json':
+        dashboard.new('My Dashboard')
+        .addTemplate(
+          {
+            current: {
+              text: 'Prometheus',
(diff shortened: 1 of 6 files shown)


brancz:
For alertmanager you can already set the cluster suffix via a flag on the prometheus operator: https://github.com/coreos/prometheus-operator/pull/3138

For the prometheus adapter I'm happy to accept this, but I think something went wrong on the commit as there are changes included that are already merged. Could you rebase this?

stafot:
Thanks for you review @brancz. I wasn't aware of this flag. 
For `prometheus-adapter`: I will rebase and let you know, as soon as possible.

stafot:
@brancz  

> When running `jb update` and `make generate` the following changes are
unstaged. If I don't add them on a commit testing pipeline fails.
Adding them in a separate commit us they are out of purpose of my
initial PR. It seems that something gone wrong on a previous commit.
My branch is in sync with `upstream/master` branch.

If you think that I did something wrong on my side I can open a fresh new PR.

brancz:
Yeah it appears that this is still the case. If you can't figure it out, I would recommend checking out latest master and doing the change for the prometheus adapter again :)

stafot:
Closed in favour of https://github.com/coreos/kube-prometheus/pull/575.

PR #576, opened 2020-06-19
Fix AlertmanagerConfigInconsistent alert

simonpasquier:
Previously the alert would fire when the number of Alertmanager pods
didn't match the number of replicas defined in the Alertmanager spec
even though all the running pods had the same configuration hash. This
type of issue is already covered by KubeStatefulSetUpdateNotRolledOut
(and possibly KubePodNotReady), having AlertmanagerConfigInconsistent
also active in this situation creates unnecessary noise.

With this change, the alert expression only returns when Alertmanager
pods have different configuration hash values irrespective of the number
of pod replicas. The message annotation has also been enhanced to report
the configuration hash for each pod.

xref: https://bugzilla.redhat.com/show_bug.cgi?id=1846397

### DIFF
--- a/jsonnet/kube-prometheus/alerts/alertmanager.libsonnet
+++ b/jsonnet/kube-prometheus/alerts/alertmanager.libsonnet
@@ -7,10 +7,15 @@
           {
             alert: 'AlertmanagerConfigInconsistent',
             annotations: {
-              message: 'The configuration of the instances of the Alertmanager cluster `{{$labels.service}}` are out of sync.',
+              message: |||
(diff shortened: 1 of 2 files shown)


brancz:
Nice!

brancz:
Is there an upstream alertmanager-mixin? If so we should probably port this over to there...

simonpasquier:
There's this long standing PR that I need to carry over: https://github.com/prometheus/alertmanager/pull/1629
It doesn't check for `alertmanager_config_hash` yet but it would be a good addition indeed.

PR #577, opened 2020-06-21
Make node-exporter listening address configurable

kradalby:
Hi

I have a few usecases were I would like to scrape the node exporter from outside of the Kubernetes cluster.

This PR makes listenAddress for node exporter configurable but continues to default to localhost to not break any current setups.


brancz:
Thanks!

Issue #545, opened 2020-05-23
Default image for prometheus operator not able to run on Raspbian / armv7


geerlingguy:
**What did you do?**

I tried deploying everything in the `manifests` directory (first the setup dir, then the parent dir), and everything deploys, but when I check my cluster, the `prometheus-operator` pod is stuck in in a `CrashLoopBackOff`.

When I deploy the Prometheus operator from [bundle.yaml](https://github.com/coreos/prometheus-operator/blob/master/bundle.yaml) directly, this issue does not occur, and the `prometheus-operator` pod launches successfully, allowing me to create Prometheus resources.

If I check the logs for the operator container in the operator pod it returns:

```
kubectl logs prometheus-operator-5cfbdc9b67-hdcwj -n monitoring -c prometheus-operator
standard_init_linux.go:211: exec user process caused "exec format error"
```

This indicates the image in use (https://github.com/coreos/kube-prometheus/blob/master/manifests/setup/prometheus-operator-deployment.yaml#L29, or quay.io/coreos/prometheus-operator:v0.39.0) doesn't have an armv7-compatible version, or maybe some other setting is in conflict that causes the operator to not launch.

**Did you expect to see some different?**

**Environment**

* Prometheus Operator version:

    `v0.5.0`

* Kubernetes version information:

    `Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.5+k3s1", GitCommit:"58ebdb2a2ec5318ca40649eb7bd31679cb679f71", GitTreeState:"clean", BuildDate:"2020-05-06T23:42:31Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/arm"}
Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.5+k3s1", GitCommit:"58ebdb2a2ec5318ca40649eb7bd31679cb679f71", GitTreeState:"clean", BuildDate:"2020-05-06T23:42:31Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/arm"}`

* Kubernetes cluster kind:

    K3s on Raspberry Pi with HypriotOS

* Manifests:

```
manifests/setup/prometheus-operator-deployment.yam
```

* Prometheus Operator Logs:

```
kubectl logs prometheus-operator-5cfbdc9b67-hdcwj -n monitoring -c prometheus-operator
standard_init_linux.go:211: exec user process caused "exec format error"
```

* Prometheus Logs:

[...]

PR #579, opened 2020-06-22
Updated prometheus adapter deployment to use a multi arch image repo

tommyjmquinn:
As per discussion here:
https://github.com/coreos/kube-prometheus/issues/545


brancz:
Thanks! :)

PR #592, opened 2020-06-30
modify kube-prometheus-static-etcd.libsonnet servicemonitorEtcd namespace

redwarn:
Prometheus does not watch  the kube-system namespce monitor object by default


brancz:
Thanks!

PR #600, opened 2020-07-03
enable etcd latency metrics in kube-apiserver

tkashem:
kube-apiserver has a histogram etcd_request_duration_seconds that measures latency between the kube-apiserver and etcd instance. This metrics is currently dropped by cluster-prometheus. Enable this metrics so we have visibility into etcd latency.

We ensured that this does not enable other unwanted metrcis

> count by(name) ({name=~"etcd_request.+"})

etcd_request_duration_seconds_bucket
etcd_request_duration_seconds_count
etcd_request_duration_seconds_sum


tkashem:
This is a screenshot of etcd latency as observed by kube-apiserver.
image

PR #610, opened 2020-07-15
Add PrometheusOperatorListErrors and fix PrometheusOperatorWatchErrors threshold 

lilic:
0.1 was a bit drastic, as currently we have a lot of errors that are benign due to the following bug https://github.com/coreos/prometheus-operator/issues/3218 The stack as the issue says, is functional but it does take an effect on the error rate. 

cc @coreos/team-monitoring

### DIFF
--- a/jsonnet/kube-prometheus/alerts/prometheus-operator.libsonnet
+++ b/jsonnet/kube-prometheus/alerts/prometheus-operator.libsonnet
@@ -5,15 +5,28 @@
         name: 'prometheus-operator',
         rules: [
           {
-            alert: 'PrometheusOperatorWatchErrors',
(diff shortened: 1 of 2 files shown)


brancz:
I think I would even be ok with starting with "all requests must be failing for a period of time" as that accurately reflects that we will not be able to make any progress in the system.

lilic:
> I think I would even be ok with starting with

@brancz can you explain a bit more what you mean here? Thanks!

brancz:
I'm saying we could even put the threshold to 100%, as the consequence is no reconcile loops are even attempted, when they should. As long as there is a non-100% rate we are still making progress and reconciling the system, so we're ok.

Issue #443, opened 2020-03-10
Does not  kube-scheduler and kube-controller-manager provide the metrics on the secure port. ?


sachinmsft:
I see that https://github.com/coreos/kube-prometheus/blob/master/manifests/prometheus-serviceMonitorKubeScheduler.yaml scrap config does not provide the option to scrap the kube-scheduler metrics on the secure port. does not kube-scheduler expose the secure metrics port ?

brancz:
They do have secure endpoints, but they didn't use to, so updating this for clusters that now expose the secure endpoints would be fantastic! :)

PR #621, opened 2020-07-25
secure metrics port for scheduler and controller-manager

tafkam:
fixes https://github.com/coreos/kube-prometheus/issues/443
tested with 1.17.9 kubeadm cluster, after kubeadm upgrade disabled the insecure metrics port for scheduler+controller.
also changed the ports for other cluster-types, but i can't really say if the secure port is provided on those platforms.


brancz:
Very nice!

This needs a re-generate via `make generate`, and the port name should probably now be `https-metrics`.

brancz:
Thank you! lgtm :tada:

PR #628, opened 2020-08-03
Pin back to master and bump prometheus-operator to 0.41.1

lilic:
cc @coreos/team-monitoring

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,7 +10,7 @@ require (
 	github.com/google/go-jsonnet v0.16.1-0.20200703153429-aaf50f5b655f
 	github.com/googleapis/gnostic v0.0.0-20170729233727-0c5108395e2d // indirect
 	github.com/imdario/mergo v0.3.7 // indirect
-	github.com/jsonnet-bundler/jsonnet-bundler v0.3.1
+	github.com/jsonnet-bundler/jsonnet-bundler v0.4.0
 	github.com/kr/pretty v0.2.0 // indirect
 	github.com/mattn/go-colorable v0.1.7 // indirect
 	github.com/pkg/errors v0.8.1
--- a/go.sum
+++ b/go.sum
@@ -39,8 +39,6 @@ github.com/golang/protobuf v1.3.2 h1:6nsPYzhq5kReh6QImI3k5qWzO4PEbvbIW2cwSfR/6xs
 github.com/golang/protobuf v1.3.2/go.mod h1:6lQm79b+lXiMfvg/cZm0SGofjICqVBUtrP5yJMmIC1U=
 github.com/google/go-cmp v0.3.1/go.mod h1:8QqcDgzrUqlUb/G2PQTWiueGozuR1884gddMywk6iLU=
 github.com/google/go-cmp v0.4.0/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
-github.com/google/go-jsonnet v0.16.0 h1:Nb4EEOp+rdeGGyB1rQ5eisgSAqrTnhf9ip+X6lzZbY0=
-github.com/google/go-jsonnet v0.16.0/go.mod h1:sOcuej3UW1vpPTZOr8L7RQimqai1a57bt5j22LzGZCw=
 github.com/google/go-jsonnet v0.16.1-0.20200703153429-aaf50f5b655f h1:mw4KoMG5/DXLPhpKXQRYTEIZFkFo0a1HU2R1HbeYpek=
 github.com/google/go-jsonnet v0.16.1-0.20200703153429-aaf50f5b655f/go.mod h1:sOcuej3UW1vpPTZOr8L7RQimqai1a57bt5j22LzGZCw=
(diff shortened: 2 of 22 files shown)


s-urbaniak:
funky, there is no prometheus being deployed :thinking:

s-urbaniak:
nevermind, i wasn't looking close enough, investigating too :thinking:

s-urbaniak:
ah, here we go
```
monitoring           prometheus-k8s-0                             2/3     InvalidImageName   0          47s
```

s-urbaniak:
```
  Warning  InspectFailed  35s (x3 over 70s)  kubelet, kind-control-plane  Failed to apply default image tag "quay.io/prometheus/prometheus:v2.20.0:v2.20.0": couldn't parse image reference "quay.io/prometheus/prometheus:v2.20.0:v2.20.0": invalid reference format
```

s-urbaniak:
yeah the `prometheus` yaml seems to duplicate the version:
```
manifests/prometheus-prometheus.yaml
14:  image: quay.io/prometheus/prometheus:v2.20.0
34:  version: v2.20.0
```

lilic:
Seems like the problem was [this PR](https://github.com/coreos/prometheus-operator/pull/3358/files) removed the way the full image is created, it is now generated from image name and tag, I fixed it in this in commit -> https://github.com/coreos/kube-prometheus/pull/628/commits/16b769c2fc3a1258c487d84cf8a44e579632602e

@brancz Before merging this, I am not sure is this want we wanted and expected from the above PR? Pinging you as you reviewed it and had some concerns on it.

brancz:
hmm no this does look like a bug in prometheus operator .. it looks like the edge case that we are hitting [here](https://github.com/coreos/kube-prometheus/pull/628/files#diff-9c5516c9bc0931ed6067ac3d7cd2addfR34) is setting all three things like this:

```
version: v2.20.0
tag: v2.20.0
image: quay.io/prometheus/prometheus:v2.20.0
```

lilic:
Yes exactly my thinking, I think if image has tag it should not be set again, or do we blame that on user error?

brancz:
note, the reasoning behind all of these changes is to make `version` less overloaded in responsibilities (in the long run) .. it was a mistake in the beginning to make it both the tag of the image as well as the "branching-factor" to decide how to provision prometheus according to the version. What I would like to see long term is making `version` only the latter, and have `image` be the only thing that decides the image.

lilic:
Agreed, I think version and tag/image should not be taken into consideration.

jsonnet/kube-prometheus/prometheus/prometheus.libsonnet
```diff
@@ -178,7 +178,8 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
         spec: {
           replicas: p.replicas,
           version: $._config.versions.prometheus,
-          image: $._config.imageRepos.prometheus + ':' + $._config.versions.prometheus,
+          image: $._config.imageRepos.prometheus,
+	  tag: $._config.versions.prometheus,
```
brancz, line 182: we should just remove this I think, then it should all work with specifying the tag in the `image`, as ultimately we want to deprecate `tag`, `sha` and `baseImage`

lilic: Done, can you have another look, thanks!

jsonnet/kube-prometheus/prometheus/prometheus.libsonnet
```diff
@@ -178,7 +178,7 @@ local k = import 'ksonnet/ksonnet.beta.4/k.libsonnet';
         spec: {
           replicas: p.replicas,
           version: $._config.versions.prometheus,
-          image: $._config.imageRepos.prometheus + ':' + $._config.versions.prometheus,
+          image: $._config.imageRepos.prometheus,
```
brancz, line 181: `image` should be the full image as it was before, if that doesn't work, then we need to fix the prometheus operator in that regard, I think it was the combination of specifying `version`, `tag`, and the full `image` that caused the edge case to trigger before

lilic: Full image with tag in the image does not work, as the version gets added to it, hence the initial failure, after removing that and adding either tag or leaving version alone that fixes it. We did not have tag before, just image and version.

brancz: Then this is broken and needs to be fixed in prometheus operator

lilic: No, sorry for not being clear enough it was just full `image` and `version`. I reverted the changes to rerun the CI again.

lilic: Yep fails, so we fix it in master and latest release before we can bump this?

brancz: yes

lilic:
Blocked by https://github.com/prometheus-operator/prometheus-operator/pull/3392

lilic:
No worries at all! Regression was fixed in prom-op 🎉

PR #634, opened 2020-08-06
jsonnet/prometheus-operator.libsonnet: Adjust alerts range

lilic:
Noticed that we have this firing for way too long, this should adjust it.


lilic:
cc @prometheus-operator/kube-prometheus-reviewers PTAL thanks!

PR #638, opened 2020-08-09
Thanos Sidecar Exposure and Scraping via ServiceMonitor

KingJ:
Expose the Thanos Sidecar's HTTP metrics and gRPC StoreAPI interfaces via a dedicated service. This allows for a new ServiceMonitor to then scrape the metrics available on the Sidecar's HTTP port. ClusterIP is set to none to allow for full discovery of all endpoints behind the service via in-cluster DNS.

A new service has been used so as to separate out the metrics that Prometheus makes available, and the metrics the Thanos Sidecar makes available. [The Thanos Mixins from thanos-io/thanos default to a job label of 'thanos-sidecar'](https://github.com/thanos-io/thanos/blob/9b578afb476374b68dd5ac5bfe50bcef70690760/mixin/config.libsonnet#L28), and hence I have applied this label to the service and specified it as the ServiceMonitor's `jobLabel`.

### DIFF
--- a/jsonnet/kube-prometheus/kube-prometheus-thanos-sidecar.libsonnet
+++ b/jsonnet/kube-prometheus/kube-prometheus-thanos-sidecar.libsonnet
@@ -26,6 +26,15 @@ local servicePort = k.core.v1.service.mixin.spec.portsType;
         ],
       },
     },
+    // Create a new service that exposes both sidecar's HTTP metrics port and gRPC StoreAPI
+    serviceThanosSidecar:
+      local thanosGrpcSidecarPort = servicePort.newNamed('grpc', 10901, 10901);
+      local thanosHttpSidecarPort = servicePort.newNamed('http', 10902, 10902);
+      service.new('prometheus-' + $._config.prometheus.name + '-thanos-sidecar', { app: 'prometheus', prometheus: $._config.prometheus.name }) +
+      service.mixin.spec.withPorts([thanosGrpcSidecarPort, thanosHttpSidecarPort]) +
+      service.mixin.spec.withClusterIp('None') +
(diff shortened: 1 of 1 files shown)


jsonnet/kube-prometheus/kube-prometheus-thanos-sidecar.libsonnet
```diff
@@ -26,6 +26,15 @@ local servicePort = k.core.v1.service.mixin.spec.portsType;
         ],
       },
     },
+    // Create a new service that exposes both sidecar's HTTP metrics port and gRPC StoreAPI
+    serviceThanosSidecar:
```
brancz, line 30: both of these should just merge into and extend the existing `Service` and `ServiceMonitor` objects

KingJ: I started doing that initially, but I wasn't able to set a different job label on the ServiceMonitor without pointing it to a dedicated service as in this case. Whilst other labels could still be used to separate out the sidecar metrics from the Prometheus metrics, I wanted to try and maintain drop in compatibility with the dashboards and alerts that the Thanos Mixin provides - which rely on the job label to uniquely identify the sidecar's metrics.

If there is a way to apply a `job=thanos-sidecar` label to metrics from an extended and unified service object though, rather than splitting them out as in this PR, then i'd be happy to implement it that way instead. I could use `jobLabel` within the ServiceMonitor to pull that from a label's value in the service itself - but semantically i'm not sure what label key to apply to the service - something like `thanos-sidecar: thanos-sidecar` would seem a little odd for example.

brancz: ok yeah makes sense, thanks for explaining! :)

brancz:
Awesome, thank you!

Issue #640, opened 2020-08-10
Quickstart servicemonitors command may be incorrect


timhughes:
**What happened?**
Following the quickstart instructions at https://github.com/prometheus-operator/kube-prometheus#quickstart
Ran the following line by line and the second line just returned immediately. 
```
kubectl create -f manifests/
until kubectl get servicemonitors --all-namespaces ; do date; sleep 1; echo ""; done
kubectl apply -f manifests/
```

When run by itself
```
thughes@silicon [0] $ kubectl get servicemonitors --all-namespaces
No resources found
thughes@silicon [0] $ echo $?
0
```
**Did you expect to see some different?**
I think i expected there to be `servicemonitors` in the output

maybe something more like the output from the following command. 
```
thughes@silicon [0] $ kubectl get crd servicemonitors.monitoring.coreos.com --all-namespaces
NAME                                    CREATED AT
servicemonitors.monitoring.coreos.com   2020-08-10T17:02:44Z
```

**How to reproduce it (as minimally and precisely as possible)**:
Follow the quickstart

**Environment**
* Prometheus Operator version:

    ```Name:                   prometheus-operator
    Namespace:              monitoring
    CreationTimestamp:      Mon, 10 Aug 2020 18:02:45 +0100
    Labels:                 app.kubernetes.io/component=controller
                            app.kubernetes.io/name=prometheus-operator
                            app.kubernetes.io/version=v0.40.0
    ```

* Kubernetes version information:

    `kubectl version`
    ```
    thughes@silicon [0] $ kubectl version
    Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.4", GitCommit:"c96aede7b5205121079932896c4ad89bb93260af", GitTreeState:"clean", BuildDate:"2020-06-17T11:41:22Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
    Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.6", GitCommit:"dff82dc0de47299ab66c83c626e08b245ab19037", GitTreeState:"clean", BuildDate:"2020-07-15T16:51:04Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
    ```

* Kubernetes cluster kind:

    kubeadmin

**Anything else we need to know?**:

brancz:
[...]

PR #641, opened 2020-08-11
Wait for CRDs to be created

timhughes:
Command now waits for all CRDs to exist before trying to use them. See #640

This isn't the best solution because if a new CRD is added then these docs will break again until updated. 

You can test it works by adding a CRD that doesn't exist which will cause it to exit with `1`

```
until kubectl get crd THIS_DOESNT_EXIST.monitoring.coreos.com alertmanagers.monitoring.coreos.com podmonitors.monitoring.coreos.com prometheuses.monitoring.coreos.com prometheusrules.monitoring.coreos.com servicemonitors.monitoring.coreos.com thanosrulers.monitoring.coreos.com ; do date; sleep 1; echo ""; done
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ Though for a quickstart a compiled version of the Kubernetes [manifests](manifes
 ```shell
 # Create the namespace and CRDs, and then wait for them to be availble before creating the remaining resources
 kubectl create -f manifests/setup
-until kubectl get servicemonitors --all-namespaces ; do date; sleep 1; echo ""; done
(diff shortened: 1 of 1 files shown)


brancz:
I think the point of the `get servicemonitors` was to make sure the apiserver is actually ready with its machinery to serve requests for the custom resource.

timhughes:
Actually i think you are right. Maybe my CRDs were created so fast that I didn't need to wait :-)

```
thughes@silicon [0] $ kubectl get non-existant-resource-type --all-namespaces
error: the server doesn't have a resource type "non-existant-resource-type"
thughes@silicon [0] $ echo $?
1
```
```
thughes@silicon [0] $ kubectl get servicemonitors --all-namespaces
No resources found
thughes@silicon [0] $ echo $?
0
```

So I think it is working as you expected and that is correct. It just isn't what I expected because I was thinking that `No resources found` meant that something was missing at that point.

timhughes:
Looks like I was over thinking things :-)

brancz:
No worries, always good to question in order to improve projects :)

PR #661, opened 2020-09-01
node-exporter: set maxUnavailable to 10%

sdodson:
This daemonset doesn't affect workload availability so allow its rollout to
be parallelized. Reduces rollout on a 250 node cluster from about 100min to 10min.


s-urbaniak:
LGTM as discussed downstream, but leaving another pair of eyes to @prometheus-operator/kube-prometheus-reviewers

brancz (approved):

PR #675, opened 2020-09-11
Use absolut jsonnet imports whenever possible

metalmatze:
There are still some dependencies that we need to make work to fully deactivate the legacyImports in the future. I'll start opening PRs against those other repositories.

~For some unknown reason, some of the rules and dashboards seem to disappear. I'll need to invest some troubleshooting time, but I'm always happy to get help.~ Already found the issue by looking at the diff here - I imported the kube-state-metrics-mixin twice instead of the node-mixin once.


brancz:
Yessssss amazing! This has been long overdue!

PR #676, opened 2020-09-13
[Fix] container_memory_working_set_bytes duplication

mohsenmottaghi:
Signed-off-by: Mohsen Mottaghi <mohsenmottaghi@outlook.com>

### Describe the bug:
since you merged helm/charts#22586 we faced with duplication on our Grafana dashboards that use container_memory_working_set_bytes metrics

### What this PR does / why we need it:
we change our queries that use container_memory_working_set_bytes for calculating memory details to use from stable /metrics/cadvisor metric path

### Which issue this PR fixes
https://github.com/helm/charts/issues/23131
https://github.com/helm/charts/issues/23249
https://github.com/helm/charts/pull/22586#issuecomment-684778852

### DIFF
--- a/manifests/grafana-dashboardDefinitions.yaml
+++ b/manifests/grafana-dashboardDefinitions.yaml
@@ -7467,7 +7467,7 @@ items:
                           "steppedLine": false,
                           "targets": [
                               {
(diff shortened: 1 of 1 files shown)


brancz:
These dashboards are defined in https://github.com/kubernetes-monitoring/kubernetes-mixin, this will need to be fixed there first :)

PR #698, opened 2020-10-02
node-exporter: Make host mounts read only and set mount propagation

brancz:
These mounts should not be writable, and we only want propagation from host to container, not the other way.


metalmatze (approved):
makes sense.

PR #736, opened 2020-10-23
Pin kube-state-metrics to last stable release branch

lilic:
Due to master going through breaking changes, using latest stable release branch release-1.9 is safest for users. Once release-2.0 branch is a bit more stable I plan on bumping it to that release.

cc @prometheus-operator/kube-prometheus-reviewers PTAL, thanks!

### DIFF
--- a/manifests/kube-state-metrics-deployment.yaml
+++ b/manifests/kube-state-metrics-deployment.yaml
@@ -3,7 +3,7 @@ kind: Deployment
 metadata:
   labels:
     app.kubernetes.io/name: kube-state-metrics
-    app.kubernetes.io/version: 1.9.7
+    app.kubernetes.io/version: v1.9.7
   name: kube-state-metrics
   namespace: monitoring
 spec:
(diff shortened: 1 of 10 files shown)


paulfantom (approved):
LGTM.

It is removing access to coordination.k8s.io/leases, but support for this is not implemented in KSM 1.9.7 so this is fine by me.

manifests/kube-state-metrics-deployment.yaml
```diff
@@ -25,8 +25,6 @@ spec:
         - --telemetry-port=8082
         image: quay.io/coreos/kube-state-metrics:v1.9.7
         name: kube-state-metrics
-        securityContext:
```
zetaab, line 28: why this was removed? Now our production does not work because of this

lilic: @zetaab Hey you can always override and use master if that works for you or release-2.0. This is just in line with what is the release-1.9 kube-state-metrics branch, feel free to raise the PR fix there against the release-1.9 branch.

zetaab: 👍  we did override for this to our jsonnet

PR #740, opened 2020-10-27
enable alternate namespace for prometheus adapter

underrun:
deleting namespaces where prometheus adapter is installed is problematic because it registers an api service there. i prefer to install to kube-system but not everyone will have the ability to do that. best option is to make namespace for prometheus adapter configurable in $._config.prometheusAdapter and default to $._config.namespace to not break current code.

### DIFF
--- a/jsonnet/kube-prometheus/kube-prometheus-custom-metrics.libsonnet
+++ b/jsonnet/kube-prometheus/kube-prometheus-custom-metrics.libsonnet
@@ -4,6 +4,7 @@
 {
   _config+:: {
     prometheusAdapter+:: {
+      namespace: $._config.namespace,
       // Rules for custom-metrics
       config+:: {
         rules+: [
@@ -87,7 +88,7 @@
       spec: {
         service: {
           name: $.prometheusAdapter.service.metadata.name,
-          namespace: $._config.namespace,
+          namespace: $._config.prometheusAdapter.namespace,
         },
         group: 'custom.metrics.k8s.io',
         version: 'v1beta1',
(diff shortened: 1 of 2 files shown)


underrun:
the workaround for this in the meantime is to alter the prometheus-adapter object comprehension to use a different variable where the namespace has been changed to what you want. modifying example.jsonnet would have something that looks like this:

```
local pa = kp +
  {
    _config+:: {
      namespace: 'kube-system',
      prometheusAdapter+:: {
        prometheusURL: 'http://prometheus-' + kp._config.prometheus.name + '.' + kp._config.namespace + '.svc.cluster.local:9090/',
      },
    },
  };
{ ['prometheus-adapter-' + name]: pa.prometheusAdapter[name] for name in std.objectFields(pa.prometheusAdapter) }
```

brancz:
I'm afraid you got caught in an on-going refactoring. Could you rebase this? Also I think the namespace should be within the `prometheusAdapter` config object.

underrun:
Rebased.

When you say:

> I think the namespace should be within the `prometheusAdapter` config object.

What do you mean if not `$._config.prometheusAdapter` - the other option would be `$.prometheusAdapter` right? If so I can make that change - but please let me know if i understand correctly.

The reason I picked `$._config.prometheusAdapter` was because the value is needed in `jsonnet/kube-prometheus/kube-prometheus-custom-metrics.libsonnet` in addition to `jsonnet/kube-prometheus/prometheus-adapter/prometheus-adapter.libsonnet` so it felt less like a local to prometheus-adapter and more like a global configuration change.

brancz:
My apologies, the PR as it is is correct. I must have looked at your comment when I was writing my previous comment.

This lgtm, thank you! :tada:

underrun:
awesome, thanks! yeah, the comment was just to help people make something work if they ran into the same issue i did.

PR #759, opened 2020-11-06
Remove ksonnet from prometheus-pvc.jsonnet

kakkoyun:


paulfantom (approved):
lgtm

Maybe in the future we should cleanup comments so they are less inlined.

brancz (approved):

PR #773, opened 2020-11-09
tests/e2e: add test to detect many-to-many errors

simonpasquier:
It happened quite a few times that some queries failed when Prometheus
scrapes metrics from 2 kube-state-metrics instances. This situation can
happen briefly when the kube-state-metrics instance is rolled out. It
might also be more apparent when automatic sharding of
kube-state-metrics is enabled.

https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/306
https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/361


brancz (approved):
Slick! I love it!

PR #778, opened 2020-11-11
install a blackbox-exporter instance

tpalfalvi:
Deploy one instance of blackbox-exporter along the other components, so we have something to execute our `Probe`s. Implements #735.

### DIFF
--- a/jsonnet/kube-prometheus/blackbox-exporter/blackbox-exporter.libsonnet
+++ b/jsonnet/kube-prometheus/blackbox-exporter/blackbox-exporter.libsonnet
@@ -106,6 +106,44 @@ local kubeRbacProxyContainer = import '../kube-rbac-proxy/container.libsonnet';
         },
       },
 
+      clusterRole: {
+        apiVersion: 'rbac.authorization.k8s.io/v1',
+        kind: 'ClusterRole',
+        metadata: {
+          name: 'blackbox-exporter',
+        },
+        rules: [
+          {
+            apiGroups: ['authentication.k8s.io'],
+            resources: ['tokenreviews'],
+            verbs: ['create'],
+          },
+          {
+            apiGroups: ['authorization.k8s.io'],
+            resources: ['subjectaccessreviews'],
+            verbs: ['create'],
+          },
+        ],
+      },
+
+      clusterRoleBinding: {
+        apiVersion: 'rbac.authorization.k8s.io/v1',
+        kind: 'ClusterRoleBinding',
+        metadata: {
+          name: 'blackbox-exporter',
+        },
+        roleRef: {
+          apiGroup: 'rbac.authorization.k8s.io',
+          kind: 'ClusterRole',
+          name: 'blackbox-exporter',
+        },
+        subjects: [{
+          kind: 'ServiceAccount',
+          name: 'blackbox-exporter',
+          namespace: $._config.namespace,
+        }],
+      },
+
       deployment: {
         apiVersion: 'apps/v1',
         kind: 'Deployment',
@@ -206,6 +244,7 @@ local kubeRbacProxyContainer = import '../kube-rbac-proxy/container.libsonnet';
           },
           spec: {
             endpoints: [{
+              bearerTokenFile: '/var/run/secrets/kubernetes.io/serviceaccount/token',
               interval: '30s',
               path: '/metrics',
               port: 'http',
--- a/jsonnet/kube-prometheus/blackbox-exporter/blackbox-exporter.libsonnet
+++ b/jsonnet/kube-prometheus/blackbox-exporter/blackbox-exporter.libsonnet
@@ -194,6 +194,7 @@
           kind: 'ServiceMonitor',
           metadata: {
             name: 'blackbox-exporter',
+            namespace: $._config.namespace,
             labels: bb.assignLabels
           },
           spec: {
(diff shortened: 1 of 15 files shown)


jsonnet/kube-prometheus/blackbox-exporter/blackbox-exporter.libsonnet
```diff
@@ -0,0 +1,211 @@
...
+                    name: 'http',
+                    containerPort: bb.port,
+                  }],
+                  resources: {
+                    requests: $._config.resources['blackbox-exporter'].requests,
+                    limits: $._config.resources['blackbox-exporter'].limits
+                  },
+                  securityContext: if bb.privileged then {
+                                     runAsNonRoot: false,
+                                     capabilities: { drop: [ 'ALL' ], add: [ 'NET_RAW'] }
```
brancz, line 134: Why would this be necessary?

brancz (commented):
Very nice! I think we should actually include this by default, not as an optional component. What do you think?

jsonnet/kube-prometheus/blackbox-exporter/blackbox-exporter.libsonnet
tpalfalvi: ICMP probes fail with permission errors if the blackbox exporter process does not have the `NET_RAW` capability.

I first added the capability to the normal unprivileged security context, but that doesn't change the effective capabilities of the processes in the container, just the allowed ones; to make it effective, I'd have to `setcap` the blackbox exporter binary, but that would mean rolling a custom container image, and I don't want to deal with that. The working solution was to run the container as root, but remove all its special privileges and only let it keep `NET_RAW`. (See e.g. prometheus/blackbox_exporter#689)

Although these privileges are quite narrow, I still didn't want to set them this way unless necessary. Folks on secure clusters may need to set up PSP to allow this; if they don't need ICMP probes, we may save them the hassle.

tpalfalvi:
By "include this by default" you mean the mixin? I kept it separate to be safe, but I don't think including it by default would break anything. Shall I make the change?

The renderer part must still be added manually, right?

brancz:
By default I meant add

```
(import 'kube-prometheus/blackbox-exporter/blackbox-exporter.libsonnet') +
```

To all the examples we have in the repo. I don't see why people wouldn't always want to install this.

brancz:
Nice this is looking good, just need to add actually rendering the blackbox exporter to the example scripts we have available.

tpalfalvi:
Yeah, but there are 34 or so example `jsonnet` files if I counted right, and probably a few other examples embedded in documentation. I'll go through them as soon as I can.

brancz:
No problem, I'd be happy with merging this, even with just the `example.jsonnet` and `examples/kustomize.jsonnet` being adapted. Those are the ones most people base their setups on, the rest are more examples of extending the base setup.

jsonnet/kube-prometheus/blackbox-exporter/blackbox-exporter.libsonnet
```diff
@@ -0,0 +1,211 @@
...
+        }
+      },
+
+      serviceMonitor:
+        {
+          apiVersion: 'monitoring.coreos.com/v1',
+          kind: 'ServiceMonitor',
+          metadata: {
+            name: 'blackbox-exporter',
+            labels: bb.assignLabels
```
peterbueschel, line 197: Thx, for your work. Just a small question here. Is there a reason, why the namespace field is missing here?

tpalfalvi: Because I wasn't paying enough attention, apparently. I must've missed it when I got rid of `ksonnet` (before posting the PR). Good catch, thank you.

tpalfalvi:
Should I put a `kube-rbac-proxy` in front of the blackbox exporter? That would take care of the e2e test failure as well.

brancz:
Yes that would be great, thank you!

tpalfalvi:
Sorry for the delay, December was a lot busier than I expected. I hope I got the RBAC thing right... at least it seems to be working.
There's another failed check now. Would rebasing the PR fix that, or is it something more serious?

brancz:
Having trouble reproducing this locally, so I re-ran the CI job.

brancz:
Ah I was able to reproduce this now. Yes, this should rebase cleanly, and a `make clean generate` and commiting the changes should do the trick. Once that's done, this is ready to merge! 

Thank you so much for bearing with us!

tpalfalvi:
Will do. Shall I squash too while I'm at it, or do you prefer it as it is?

brancz:
I'm happy with the history as is.

brancz:
Awesome, thank you! :tada:

brancz:
I was just trying this out, and looks like we overlooked something.. while we can successfully scrape the blackbox exporter's own metrics via kube-rbac-proxy, it's not possible to set TLS and authn/authz settings for the Prober in the ProbeSpec (yet). I'll open a patch that makes the blackbox exporter bind against all interfaces again, until we implement the missing functionality in the prometheus operator for this.

brancz:
Turns out the change was actually minimal and you already took care of almost everything! :tada:

Thank you so much for this contribution, works wonderfully!

tpalfalvi:
The weird thing is, it seems to be working in my cluster. My Prometheus is getting fresh HTTP/TCP/... scrape results from my blackbox exporter instance. I'm looking into why it works when it shouldn't...

tpalfalvi:
... it was sending probes to the internal port for some reason. Maybe the service was pointing there for a short time while I was hacking around, and it stuck due to some keepalive setting? Deleting and re-adding all `Probe`s fixed the anomaly (i.e. broke all the blackbox scrapes). Sorry for missing this.

brancz:
No worries, as I said it was really just a minuscule thing, very nice work!

PR #781, opened 2020-11-12
Revert kubeadm selector and replace nodePort

philschleier:
While I don't use the kubeadm, I try to keep my Kubernetes deployments compatible with it. During some ksonnet removal the selector was recently changed from component to k8s-app, which I think was accidental. (https://github.com/prometheus-operator/kube-prometheus/pull/764)

As far as I can tell, kubeadm still uses the component label: https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/util/staticpod/utils.go#L63

Also the node-ports mixin now adds a second port with the same name, so I end up with two ports with same name and targetPort, but one of them defining the nodePort. I don't think this was intentional either.

This PR reverts the label for kubeadm to be 'component' and the node-ports mixin to replace the port instead of add a second one.


kakkoyun:
@philschleier Thanks for fixing my mess 🙈

PR #800, opened 2020-11-20
Provide external-metrics as an add-on

kehao95:
In (#374) we include `custom-metrics` as an add-on. 
Given the fact that doc for `external-metrics` has improved recently [#302](https://github.com/DirectXMan12/k8s-prometheus-adapter/pull/302). I think we can also provide external-metrics as well. 
Before the performance issue is addressed we just keep it as an add-on.
> yes, agreed. I share the same assertion that prometheus-adapter is not production ready yet regarding custom and external metrics.
_Originally posted by @s-urbaniak in https://github.com/prometheus-operator/kube-prometheus/issues/374#issuecomment-573692256_

> The design started it's live here https://github.com/s-urbaniak/prometheus-adapter/blob/master/design.md, along side with an initial implementation (that actually also booted up).
The configmap proposal in that document for resource metrics is actually something I envisioned (aka a nearly zero-config expression of a sample query without templating)
The CRD design for custom and external metrics is TBD, but would follow the same idea above.
_Originally posted by @s-urbaniak in https://github.com/prometheus-operator/kube-prometheus/issues/374#issuecomment-575592123_

I am currently using it in production and looks good so far.

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,6 +208,7 @@ local kp =
   // (import 'kube-prometheus/kube-prometheus-static-etcd.libsonnet') +
   // (import 'kube-prometheus/kube-prometheus-thanos-sidecar.libsonnet') +
   // (import 'kube-prometheus/kube-prometheus-custom-metrics.libsonnet') +
+  // (import 'kube-prometheus/kube-prometheus-external-metrics.libsonnet') +
   {
     _config+:: {
       namespace: 'monitoring',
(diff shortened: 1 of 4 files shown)


s-urbaniak:
This LGTM, thank you so much!

Side note: i am wondering how we can make sure this does not break over time. We have simple e2e tests here but (unless I miss something) these do not include optional assets, cc @prometheus-operator/kube-prometheus-reviewers.

s-urbaniak:
@kehao95 when trying out locally I get the following though:
```
v1beta1.external.metrics.k8s.io        monitoring/prometheus-adapter   False (FailedDiscoveryCheck)   4m
v1beta1.metrics.k8s.io                 monitoring/prometheus-adapter   True                           4m
```

s-urbaniak:
I didn't dig deeper but there seems to be something wrong with the manifests.

s-urbaniak:
Ah sorry, my bad, I forgot to uncomment the example configuration, as that did not have enabled the external endpoint from prometheus-adapter.

Thank you, merging!

PR #821, opened 2020-12-04
Add GKE patch

brancz:
Follow up from https://github.com/prometheus-operator/prometheus-operator/issues/2211

GKE enforces `metrics-server` to be used and there is nothing we can do against that, so for GKE we should disable the resource metrics API.

cc @lilic @paulfantom @simonpasquier 

Tested on @polarsignals prod cluster :wink:

### DIFF
--- a/jsonnet/kube-prometheus/kube-prometheus-gke.libsonnet
+++ b/jsonnet/kube-prometheus/kube-prometheus-gke.libsonnet
@@ -1,0 +1,13 @@
+(import './kube-prometheus-managed-cluster.libsonnet') + {
+  _config+:: {
+    prometheusAdapter+:: {
+      config+: {
+        resourceRules:: null,
+      },
+    },
+  },
+
+  prometheusAdapter+:: {
+    apiService:: null,
+  },
+}


lilic (approved):
❄️ 🧊 
lgtm Thanks!

coderanger:
It is technically possible to disable GKE's monitoring stack, though anyone that advanced probably can figure out how to deal with this themselves :)

brancz:
Yeah, I feel for most users it's not worth the fight though.

coderanger:
This is probably the best way, only one thing can register against metrics.k8s.io at a time. We could still install stuff on a different API name but most metrics API consumers are hardwired to the specific names :-( Probably a thing to take up with sig-api-machinery since it's definitely a place that hurts operator lifecycle management in general. Keda has similar issues.

brancz:
Yeah, I've also thought of at least allowing different providers for different namespaces.

coderanger:
Or just getting people in a pattern of allowing the consumer to configure their metrics API group more flexibly. Like HPA's metrics struct could have an optional string to override the API group and then we could document using that or whatever. Still hard to get a good, modular UX but at least fewer hard conflicts.

PR #828, opened 2020-12-09
Cut release-0.7

dgrisonnet:
Cut release 0.7 of kube-prometheus.

Once merged, @s-urbaniak will finalize the release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ This adapter is an Extension API Server and Kubernetes needs to be have this fea
 To try out this stack, start [minikube](https://github.com/kubernetes/minikube) with the following command:
 
 ```shell
-$ minikube delete && minikube start --kubernetes-version=v1.19.0 --memory=6g --bootstrapper=kubeadm --extra-config=kubelet.authentication-token-webhook=true --extra-config=kubelet.authorization-mode=Webhook --extra-config=scheduler.address=0.0.0.0 --extra-config=controller-manager.address=0.0.0.0
+$ minikube delete && minikube start --kubernetes-version=v1.20.0 --memory=6g --bootstrapper=kubeadm --extra-config=kubelet.authentication-token-webhook=true --extra-config=kubelet.authorization-mode=Webhook --extra-config=scheduler.address=0.0.0.0 --extra-config=controller-manager.address=0.0.0.0
 ```
 
 The kube-prometheus stack includes a resource metrics API server, so the metrics-server addon is not necessary. Ensure the metrics-server addon is disabled on minikube:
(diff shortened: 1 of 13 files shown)


README.md
```diff
@@ -94,19 +94,20 @@ $ minikube addons disable metrics-server
...
-| `release-0.3`         | ✔               | ✔               | ✔               | ✔               | ✗               | ✗               |
-| `release-0.4`         | ✗               | ✗               | ✔ (v1.16.5+)    | ✔               | ✗               | ✗               |
-| `release-0.5`         | ✗               | ✗               | ✗               | ✗               | ✔               | ✗               |
-| `release-0.6`         | ✗               | ✗               | ✗               | ✗               | ✔               | ✔               |
-| `HEAD`                | ✗               | ✗               | ✗               | ✗               | x               | ✔               |
+| kube-prometheus stack | Kubernetes 1.14 | Kubernetes 1.15 | Kubernetes 1.16 | Kubernetes 1.17 | Kubernetes 1.18 | Kubernetes 1.19 | Kubernetes 1.20 |
+|-----------------------|-----------------|-----------------|-----------------|-----------------|-----------------|-----------------|-----------------|
+| `release-0.3`         | ✔               | ✔               | ✔               | ✔               | ✗               | ✗               | ✗               |
```
dgrisonnet, line 99: Should we keep supporting `release-0.3`?

s-urbaniak:
cc @prometheus-operator/kube-prometheus-reviewers

README.md
lilic: Lets remove the mention of release-0.3 yes, we will still support it. This is compatibility rather than support scheme.

I would stick to last 5 releases (including master) and last 5 k8s versions.

jsonnet/kube-prometheus/jsonnetfile.json
```diff
@@ -44,7 +44,7 @@
           "subdir": ""
         }
       },
-      "version": "master"
+      "version": "release-0.6"
```
lilic, line 47: These we can just do as a PR to the release branch after its cut, otherwise here in master you need to undo the changes, as we always want master to point to master.

lilic (commented):
Thanks!

I would also like us to bump k8s version to test against 1.20 as well, should be out now. WDYT?

dgrisonnet:
@lilic I already updated the CI to test against both 1.19 and 1.20, did you had anything else in mind?

jsonnet/kube-prometheus/jsonnetfile.json
dgrisonnet: @lilic I guess I should also do the same for kube-state-metrics?

lilic:
No, just did not show up for me before, all good, thanks! 👍 🎉

jsonnet/kube-prometheus/jsonnetfile.json
lilic: kube-state-metrics alerts mixin is okay to be master or pinned, either work as we did not change anything there, but we can keep it pinned to 1.9 as well. Important is to not change ksm own mixin to master, as we have breaking changes in master.

README.md
dgrisonnet: Ok, I'll remove the 1.14 and 1.15 rows. However, since we don't have any version fully compatible with 1.16, is it meaningful to keep release-0.3 for now?

lilic: > Lets remove the mention of release-0.3 yes, we will still support it. 
> I would stick to last 5 releases (including master) and last 5 k8s versions.

Yes as mentioned, I believe 5 versions master being one of them, removes the release-0.3 no?

jsonnet/kube-prometheus/jsonnetfile.json
```diff
@@ -35,7 +35,7 @@
           "subdir": "jsonnet/mixin"
         }
       },
-      "version": "master"
+      "version": "release-0.44"
```
lilic, line 38: Sorry for not pointing out, this can also stay at master and can just be done as a PR towards the release branch, otherwise we need to change it twice as in master it will always stay master. If that makes sense?

dgrisonnet: Makes sense, sorry I misunderstood your first comment about this. It should be good now.

dgrisonnet:
The E2E tests seems to be failing because prometheus-operator v0.44.1 has not been fully released yet.

lilic (approved):
lgtm on green, thanks!

PR #830, opened 2020-12-10
Pin jsonnet dependencies in release-0.7

dgrisonnet:
Pin the following jsonnet mixins for the 0.7.0 release:
- prometheus-operator mixins: release-0.44
- kube-state-metrics mixins: release-1.9
- kubernetes-mixins: release-0.6
- kubernetes-grafana: release-0.2

/cc @s-urbaniak @lilic

### DIFF
--- a/manifests/grafana-dashboardDefinitions.yaml
+++ b/manifests/grafana-dashboardDefinitions.yaml
@@ -16657,7 +16657,7 @@ items:
                           "tableColumn": "",
                           "targets": [
                               {
-                                  "expr": "sum(kubelet_running_pods{cluster=\"$cluster\", job=\"kubelet\", metrics_path=\"/metrics\", instance=~\"$instance\"}) OR sum(kubelet_running_pod_count{cluster=\"$cluster\", job=\"kubelet\", metrics_path=\"/metrics\", instance=~\"$instance\"})",
+                                  "expr": "sum(kubelet_running_pods{cluster=\"$cluster\", job=\"kubelet\", metrics_path=\"/metrics\", instance=~\"$instance\"})",
                                   "format": "time_series",
                                   "intervalFactor": 2,
                                   "legendFormat": "{{instance}}",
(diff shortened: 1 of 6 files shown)


s-urbaniak:
lgtm on green

manifests/grafana-dashboardDefinitions.yaml
```diff
@@ -16657,7 +16657,7 @@ items:
                           "tableColumn": "",
                           "targets": [
                               {
-                                  "expr": "sum(kubelet_running_pods{cluster=\"$cluster\", job=\"kubelet\", metrics_path=\"/metrics\", instance=~\"$instance\"}) OR sum(kubelet_running_pod_count{cluster=\"$cluster\", job=\"kubelet\", metrics_path=\"/metrics\", instance=~\"$instance\"})",
+                                  "expr": "sum(kubelet_running_pods{cluster=\"$cluster\", job=\"kubelet\", metrics_path=\"/metrics\", instance=~\"$instance\"})",
```
s-urbaniak, line 16660: out of curiosity: why is this different than on master?

lilic (commented):
There are some differences to the previous release, any reason for not pinning grafana version same as with last release? https://github.com/prometheus-operator/kube-prometheus/blob/release-0.6/jsonnet/kube-prometheus/jsonnetfile.json

manifests/grafana-dashboardDefinitions.yaml
dgrisonnet: this is a recent change that is only in kubernetes-mixins master branch and not on the release-0.6 one.

dgrisonnet:
I missed it, thank you for pointing it out!

lilic:
No problem, I was not sure if we want a release of that or pin to a commit or if much changed. But I remember we pinned last time due to some problems we had after an update on a branch when it was pinned to master.

manifests/grafana-dashboardDefinitions.yaml
lilic: We might have to consider bringing this to the newest mixin branch if its a fix. But we can do that outside of this PR.

lilic (approved):
lgtm

PR #858, opened 2021-01-04
kubelet: Revert label selector

brancz:
The Kubelet's Service/Endpoints object maintained by the Prometheus
Operator does not have the recommended app label (yet). Therefore we
need to use the old label until a Prometheus Operator version has been
released and integrated in kube-promteheus that does use it.

Related to https://github.com/prometheus-operator/prometheus-operator/pull/3768


paulfantom (approved):
lgtm

PR #886, opened 2021-01-22
feat(Grafana-Dashboards): add cluster label to some metrics 

secustor:
Some queries in Grafana dashboards are missing the cluster labels. 

This PR tries to fix this.

### DIFF
--- a/manifests/grafana-dashboardDefinitions.yaml
+++ b/manifests/grafana-dashboardDefinitions.yaml
@@ -7547,7 +7547,7 @@ items:
                           "steppedLine": false,
                           "targets": [
                               {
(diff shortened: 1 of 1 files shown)


paulfantom (changes requested):
Those dashboards are coming from https://github.com/kubernetes-monitoring/kubernetes-mixin project and should be adjusted there. After this is completed we can update the project dependency.

secustor:
Here is the PR at the upstream project https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/542

vsxen:
I can not find any cluster label in those metrics, such as `container_memory_working_set_bytes `。why?

secustor:
I have the cluster label at these metrics: 
image

vsxen:
do you set external_labels in prom?

https://github.com/prometheus-operator/kube-prometheus/search?l=YAML&p=1&q=cluster

paulfantom:
This seems to be already completed. Please reopen if this is not already included.

PR #900, opened 2021-01-29
Default alerting to Alertmanager API v2

simonpasquier:
Alertmanager API v2 is available for more than 2 years now, there's no
reason to not use it by default.

cc @paulfantom 

see also https://github.com/prometheus/alertmanager/issues/2469 and https://github.com/prometheus/prometheus/issues/8398


lilic (approved):
Thanks! lgtm

PR #912, opened 2021-02-03
Include build.sh and example.jsonnet download

aisbaa:
I've tripped on this while trying to compile kube-prometheus manifests. On master branch instructions point to release-0.4 which is not compatible with build.sh and example.jsonnet from master branch.

Ref: https://github.com/prometheus-operator/kube-prometheus/issues/902


brancz:
Nice!

PR #956, opened 2021-02-19
Ensure Prometheus ServiceMonitor is unique

maxbrunet:
The name of the Prometheus ServiceMonitor must be unique like other resources to allow deploying multiple "shards" in the same namespace.
This has been tested, we have be deploying with this patch for quite some time now.


paulfantom (approved):
lgtm, thanks! :+1:

PR #1057, opened 2021-03-24
Update kubernetes mixins

jan--f:


lilic:
Thanks for the PR! Seems odd that the CI is failing. Also we should really bump to test against the 1.21 k8s version. I rerun the jobs lets see if its just a flake.

jan--f:
Just a flake than I guess?

lilic (approved):
lgtm

Seems like a flake yes. Thanks for the bump!

PR #1060, opened 2021-03-25
PodSecurityPolicy uses role instead of clusterRole where posible

ArthurSens:
Following up some post-merge comments at #1010


lilic:
Thank you for opening this!

cc @brancz I think you were the one who had the comments.

brancz:
Nice, thank you! lgtm :+1: :tada:

PR #1063, opened 2021-03-29
add cluster role to list and watch ingresses in "networking.k8s.io"

viperstars:
I found the prometheus cannot list and watch ingresses in api group "networking.k8s.io".

The Ingress I created is:
```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: prometheus
  namespace: monitoring
spec:
  rules:
  - host: prometheus.minikube
    http:
      paths:
      - pathType: Prefix
        path: "/"
        backend:
          service: 
            name: prometheus-k8s
            port: 
              number: 9090
```

And error log is:
```
level=error ts=2021-03-26T15:58:34.357Z caller=klog.go:96 component=k8s_client_runtime func=ErrorDepth msg="pkg/mod/k8s.io/client-go@v0.20.2/tools/cache/reflector.go:167: Failed to watch *v1beta1.Ingress: failed to list *v1beta1.Ingress: ingresses.networking.k8s.io is forbidden: User \"system:serviceaccount:monitoring:prometheus-k8s\" cannot list resource \"ingresses\" in API group \"networking.k8s.io\" in the namespace \"monitoring\""
```

After I add proper role for the prometheus manually, it works.

So I update the jsonnet file and generate all yaml files in "manifests".

### DIFF
--- a/jsonnet/kube-prometheus/components/prometheus.libsonnet
+++ b/jsonnet/kube-prometheus/components/prometheus.libsonnet
@@ -229,6 +229,11 @@ function(params) {
           resources: ['ingresses'],
           verbs: ['get', 'list', 'watch'],
         },
+        {
+          apiGroups: ['networking.k8s.io'],
+          resources: ['ingresses'],
+          verbs: ['get', 'list', 'watch'],
+        },
       ],
     };
     {
--- a/manifests/prometheus-roleSpecificNamespaces.yaml
+++ b/manifests/prometheus-roleSpecificNamespaces.yaml
@@ -29,6 +29,14 @@ items:
     - get
     - list
     - watch
+  - apiGroups:
+    - networking.k8s.io
+    resources:
+    - ingresses
+    verbs:
+    - get
+    - list
+    - watch
 - apiVersion: rbac.authorization.k8s.io/v1
   kind: Role
   metadata:
@@ -58,6 +66,14 @@ items:
     - get
     - list
     - watch
+  - apiGroups:
+    - networking.k8s.io
+    resources:
+    - ingresses
+    verbs:
+    - get
+    - list
+    - watch
 - apiVersion: rbac.authorization.k8s.io/v1
   kind: Role
   metadata:
@@ -81,6 +97,14 @@ items:
     - watch
   - apiGroups:
     - extensions
+    resources:
+    - ingresses
+    verbs:
+    - get
+    - list
+    - watch
+  - apiGroups:
+    - networking.k8s.io
     resources:
     - ingresses
     verbs:


lilic:
Hey, thanks for the PR! Just one question, when do we watch Ingress resource of the box, or is this something you added to your own setup?

viperstars:
> Hey, thanks for the PR! Just one question, when do we watch Ingress resource of the box, or is this something you added to your own setup?



>The v1.22 release will stop serving the following deprecated API versions in favor of newer and more stable API versions:
>   Ingress in the extensions/v1beta1 API version will no longer be served
>   Migrate to use the networking.k8s.io/v1beta1 API version, available since v1.14. Existing persisted data can be retrieved/updated via the new version.

[Source](https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/)

There is a API Migration for ingress. Before kubernetes v1.22, we can use both "extensions/v1beta1" and "networking.k8s.io/v1beta1" as the apiVersion.

If user apply ingress with apiVersion "extensions/v1beta1", the watch will be successful, because of existing cluster role: 
```
        {
          apiGroups: ['extensions'],
          resources: ['ingresses'],
          verbs: ['get', 'list', 'watch'],
        },
```

But if user apply ingress with apiVersion "networking.k8s.io/v1beta1" or "networking.k8s.io/v1", the watch will not be successful, because there is no proper cluster role.

This pr is to add cluster role for ingress with apiVersion "networking.k8s.io".
```
        {
          apiGroups: ['networking.k8s.io'],
          resources: ['ingresses'],
          verbs: ['get', 'list', 'watch'],
        },
```

viperstars:
> Hey, thanks for the PR! Just one question, when do we watch Ingress resource of the box, or is this something you added to your own setup?

If we create an ingress with api version "networking.k8s.io/v1":
```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  labels:
    probe: enabled
  name: prometheus
  namespace: monitoring
spec:
  rules:
  - host: prometheus.minikube
    http:
      paths:
      - pathType: Prefix
        path: "/"
        backend:
          service: 
            name: prometheus-k8s
            port: 
              number: 9090
```

And then create a probe for the ingresses:
```yaml
apiVersion: monitoring.coreos.com/v1
kind: Probe
metadata:
  labels:
    app.kubernetes.io/component: exporter
  name: probe-ingress
  namespace: monitoring
spec:
  module: http_2xx
  prober:
    url: blackbox-exporter.monitoring.svc:19115
  targets:
    ingress:
      namespaceSelector:
        any: true
      selector:
        matchLabels:
          probe: enabled
```

This will make the promtheus instance which created by the operator watch ingresses with api version "networking.k8s.io/v1".

lilic (approved):
Makes sense Probes, thank you for the PR!

PR #1070, opened 2021-03-30
Psp should be deployed at the same namespace as kube-prometheus stack

ArthurSens:
@brancz I found out the issue on my end, and I think the fix makes sense upstream as well :)


paulfantom (approved):
lgtm

PR #1074, opened 2021-04-02
Setup gitpod config for ephemeral dev-environments

ArthurSens:
This PR adds configuration that makes life easier if you want to develop this project on [Gitpod](https://www.gitpod.io/) like myself 😛.

It installs all the necessary tooling using the Makefile script then adds `./tmp/bin` to `PATH` so we can use all those binaries.

It also installs a jsonnet extension with syntax highlighting and autocompletion:

image

You can test the difference by accessing how things are today on main:
https://gitpod.io/#https://github.com/prometheus-operator/kube-prometheus

And then compare with how things are on this branch:
https://gitpod.io/#https://github.com/ArthurSens/kube-prometheus/tree/as/gitpodfy

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,0 +1,8 @@
+
+tasks: 
+  - init: |
+      make --always-make
+      export PATH="$(pwd)/tmp/bin:${PATH}"
+vscode:
+  extensions:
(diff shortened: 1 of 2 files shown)


paulfantom (approved):
Awesome!

kakkoyun (approved):
LGTM 🚀

PR #1148, opened 2021-05-17
jsonnet/components: fix missing resource config in blackbox exporter

xadereq:
There was missing resources configuration for kubeRbacProxy container in blackbox exporter deployment.
Because of that, I wasn't able to change all resource limits using config value override.


xadereq:
@paulfantom

paulfantom (approved):
lgtm on gren

PR #1156, opened 2021-05-25
Update PodDisruptionBudget version to policy/v1

fpetkovski:
The policy/v1beta1 api version is deprecated in Kubernetes 1.21 and
will be removed in 1.25.

This commit updates the api version for all PodDisruptionBudgets
to policy/v1.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,8 @@ JSONNETFMT_ARGS=-n 2 --max-blank-lines 2 --string-style s --comment-style s
 MDOX_VALIDATE_CONFIG?=.mdox.validate.yaml
 MD_FILES_TO_FORMAT=$(shell find docs developer-workspace examples experimental jsonnet manifests -name "*.md") $(shell ls *.md)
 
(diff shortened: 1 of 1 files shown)


fpetkovski:
We need to wait with merging this PR until kube-prometheus stops supporting Kubernetes 1.20.

dgrisonnet:
From a supportability point of view, dropping 1.20 in main should be fine since there are already other releases of the project covering this version. 
That said, this change will definitely break users currently using the main branch of kube-prometheus on 1.20. In that case, they could either fall back to release-0.8 or upgrade their cluster to 1.21, but I wouldn't expect the project to be blocking improvements on main because of supportability. wdyt @paulfantom?

lilic:
The version might be deprecated but it's not removed yet, so if we can hold off a bit more, agreed with @paulfantom here.

fpetkovski:
Looks like the changeset is empty after rebasing against main. Safe to close?

PR #1160, opened 2021-05-26
Test against Kubernetes v1.21

dgrisonnet:
According to the compatibility matrix, the main branch supports both Kubernetes 1.20 and 1.21, so we should run the tests on both versions.


dgrisonnet:
I am planning on backporting this change to `release-0.8` as it also supports 1.21.

dgrisonnet:
Closing in favor of #1161

PR #1193, opened 2021-06-08
Update alertmanager mixin

ArthurSens:
Bumping Alertmanager mixin version, and adding the dashboard developed in: https://github.com/prometheus/alertmanager/pull/2540

cc/ @paulfantom


ArthurSens:
Without making any changes, this is what I see after applying the generated manifests 🙂:

image

paulfantom:
Awesome!!! :100:

PR #1211, opened 2021-06-15
[Gitpod] Deploy kube-prometheus on k3s

ArthurSens:
Based on this awesome example: https://github.com/fntlnz/gitpod-k3s

---

This PR automates the deployment of kube-prometheus into a k3s cluster, running inside a VM, running inside a Gitpod Workspace.

The aim of this PR is to enable easy visualization of dashboards and alerts when updating upstream dependencies,  without requiring the reviewer to start a cluster locally. 

The experience right now is not great, Gitpod limits CPU for Workspaces that consume huge amounts, and running a VM that is able to support a k3s cluster is not exactly the lightest thing 😅. After opening the workspace, it takes around 20~30 mins to spin up VM, k3s and deploy kube-prometheus. Navigating through Grafana dashboards can be quite sluggish and I see a couple of alerts firing due to that same sluggishness 😞.

While waiting 20mins to be able to see a Grafana dashboard is not the best, this whole process can be further automated with [Gitpod Prebuilds](https://www.gitpod.io/docs/prebuilds/), unfortunately, I wasn't able to make k3s work with Prebuilds for our own internal repositories. You can expect another PR in the future to eliminate this 20~30min wait time once I get that working 🙂.

Quick video explaining how to visualize dashboards and alerts:
https://www.loom.com/share/0a4a6ceca9cb450d889811e3ec41192e

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@ vendor/
 ./auth
 .swp
 crdschemas/
+
+.gitpod/_output/
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,4 +1,5 @@
-
+image: gitpod/workspace-full
+checkoutLocation: gitpod-k3s
 tasks: 
   - init: |
       make --always-make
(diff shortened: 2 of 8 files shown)


paulfantom (approved):
THIS IS AWESOME!!! :100:

PR #1212, opened 2021-06-16
Fix ingress path rules for networking.k8s.io/v1

sanglt:
Current ingress path rules are not working. This PR update the path rules to make sure it working with apiVersion networking.k8s.io/v1.


paulfantom (approved):
lgtm on green. Thanks! :100:

ArthurSens:
Hi @sanglt,

You can make the CI happy by just running `make fmt` or `make fmt --always-make` and commit the changes 🙂

ArthurSens:
Thanks!

PR #1227, opened 2021-06-25
.github/workflows: Update versions schedule to run each Monday

fpetkovski:
Updates the versions.yaml workflow to run each Monday instead of each day.


dgrisonnet (approved):
Awesome, thanks!

cc @prometheus-operator/kube-prometheus-reviewers for awareness

PR #1233, opened 2021-07-01
Fix make manifests not building every time

sthaha:
Make target `manifests` has a dependency on build.sh which if untouched
wouldn't generate the manifests after the first run. This patch fixes it
by removing the `build.sh` dependency

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ generate: manifests **.md
 **.md: $(EMBEDMD_BIN) $(shell find examples) build.sh example.jsonnet
 	$(EMBEDMD_BIN) -w `find . -name "*.md" | grep -v vendor`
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -28,15 +28,15 @@ generate: manifests **.md
...
+manifests: examples/kustomize.jsonnet $(GOJSONTOYAML_BIN) vendor
 	./build.sh $<
 
 vendor: $(JB_BIN) jsonnetfile.json jsonnetfile.lock.json
 	rm -rf vendor
 	$(JB_BIN) install
 
 crdschemas: vendor
-	./scripts/generate-schemas.sh	
+	./scripts/generate-schemas.sh
```
sthaha, line 39: Unrelated change but please allow this cleanup 🙏

PR #1234, opened 2021-07-01
sync: Update 0.8 kube-mixin dependencies and generate

PhilipGough:
Sync upstream dependencies via `jb update`. The driver behind this sync is https://bugzilla.redhat.com/show_bug.cgi?id=1978208 where dashboards fail to show pod name and instead show {{pod}}. The fix was back ported in https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/618

### DIFF
--- a/jsonnetfile.lock.json
+++ b/jsonnetfile.lock.json
@@ -59,8 +59,8 @@
           "subdir": ""
         }
       },
-      "version": "ec3e85f45b5691d54a02ab38ed654c3c9f736fe5",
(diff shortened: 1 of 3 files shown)


dgrisonnet:
I'd prefer if we only update the dependency that contains the fix your need to backport, which I presume is kubernetes-mixins. Also, would you mind detailing in the description of this PR, what is the fix you are trying to backport?

paulfantom (approved):
lgtm

PhilipGough:
Screenshot 2021-07-02 at 15 14 34

Pods name seems to be showing correctly after this change

PR #1282, opened 2021-07-27
Add resource requests and limits to prometheus-adapter container

lanmarti:
## Description
While testing out release 0.8 I noticed there were no container requests/limits for prometheus-adapter. Looking at the jsonnet file, I saw that resources were defined in the default values, but not added to the container. By adding the line in the PR, resource limits are now attached to the container.
I have to note that the resource requests and limits are a lot higher than the usage I'm seeing in my cluster, but I'm running a small cluster. Still, if anyone could verify that the actual resources defined are reasonable, that would be great!

## Type of change

## Changelog entry

```release-note
- Add resource requests and limits to prometheus-adapter container
```


paulfantom (approved):
Great catch! Thank you :100:

PR #1308, opened 2021-08-02
Create Thanos Sidecar rules separately from Prometheus ones

maxbrunet:
## Description

I want to be able to hide these resources easily:

```jsonnet
    prometheusRuleThanosSidecar:: {},
    serviceThanosSidecar:: {},
    serviceMonitorThanosSidecar:: {},
```

I prefer to manage them with https://github.com/thanos-io/kube-thanos/blob/main/jsonnet/kube-thanos/kube-thanos-sidecar.libsonnet instead, it can be used to create a single set of resources for all Prometheus pairs, which avoids updating Thanos Query for each of them.

I also think PrometheusRule could create duplicated rules between Prometheus pairs, as the Prometheus query selector does not exclusively select the sidecar metrics it has been deployed for.

(And it happens to be a bit more readable too)

## Type of change

## Changelog entry

```release-note
- Create Thanos Sidecar rules separately from Prometheus ones
```

### DIFF
--- a/jsonnet/kube-prometheus/components/prometheus.libsonnet
+++ b/jsonnet/kube-prometheus/components/prometheus.libsonnet
@@ -59,6 +59,7 @@ function(params) {
   mixinThanos::
     (import 'github.com/thanos-io/thanos/mixin/alerts/sidecar.libsonnet') +
     (import 'github.com/kubernetes-monitoring/kubernetes-mixin/lib/add-runbook-links.libsonnet') + {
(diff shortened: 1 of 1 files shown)


paulfantom (approved):
lgtm

@dgrisonnet please take a look as this might have an impact on cluster-monitoring-operator.

dgrisonnet:
lgtm, thank you for the heads-up @paulfantom

PR #1344, opened 2021-08-20
jsonnet: Support scraping the config-reloader for AlertManager and Pr…

PhilipGough:
…ometheus

## Description

This change updates the ServiceMonitor to scrape the reloader sidecar for Promtheus and AlertManager via the service when the components are not set to listen on localhost.

## Type of change

## Changelog entry

```release-note
Support scraping config-reloader sidecar for Prometheus and AlertManager StatefulSets
```

### DIFF
--- a/jsonnet/kube-prometheus/components/alertmanager.libsonnet
+++ b/jsonnet/kube-prometheus/components/alertmanager.libsonnet
@@ -19,6 +19,7 @@ local defaults = {
     if !std.setMember(labelName, ['app.kubernetes.io/version'])
   },
   name: error 'must provide name',
+  reloaderPort: 8080,
   config: {
     global: {
       resolve_timeout: '5m',
@@ -136,6 +137,7 @@ function(params) {
     spec: {
       ports: [
         { name: 'web', targetPort: 'web', port: 9093 },
+        { name: 'reloader-web', port: am._config.reloaderPort, targetPort: 'reloader-web' },
       ],
       selector: {
         app: 'alertmanager',
@@ -161,6 +163,7 @@ function(params) {
       },
       endpoints: [
         { port: 'web', interval: '30s' },
+        { port: 'reloader-web', interval: '30s' },
       ],
     },
   },
--- a/jsonnet/kube-prometheus/components/prometheus.libsonnet
+++ b/jsonnet/kube-prometheus/components/prometheus.libsonnet
@@ -35,6 +35,7 @@ local defaults = {
     },
   },
   thanos: null,
+  reloaderPort: 8080,
 };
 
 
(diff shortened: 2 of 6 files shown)


ArthurSens (approved):
Just tested and it works as expected :) 
image


Looks like reloader only adds 7 metrics, pretty cheap. From my side, it is ok to have the scrape enabled by default, but I'll let others leave their opinion prior to merging

PhilipGough:
@ArthurSens thanks for the review and testing it. I was happy to leave it off as default actually but just wanted to ensure CI passed with it enabled. As you said, we can wait to see what others think and am happy to leave it or change it either way.

jsonnet/kube-prometheus/main.libsonnet
```diff
@@ -19,6 +19,8 @@ local utils = import './lib/utils.libsonnet';
     common: {
       namespace: 'default',
       platform: null,
+      scrapeConfigReloader: true,
```
dgrisonnet, line 22: I don't think this should be conditional since we thrive to monitor all the applications that we run.

jsonnet/kube-prometheus/main.libsonnet
```diff
@@ -19,6 +19,8 @@ local utils = import './lib/utils.libsonnet';
...
+      reloaderPort: 8080,
```
dgrisonnet, line 23: This value is specific to each component so I wouldn't define its value globally but rather per component.

dgrisonnet (commented):
Just a few comments from my side but it looks great overall, thanks @PhilipGough.

jsonnet/kube-prometheus/main.libsonnet
PhilipGough: @dgrisonnet - understood, my thinking was that there might be situations where the downstream might be injecting a proxy for example or setting the reloader to listen on localhost where it wouldn't be scrape-able via the service monitor. Any concerns in that respect?

PhilipGough:
@dgrisonnet removed the option for conditional config and added component specific config for the port as requested.

jsonnet/kube-prometheus/components/alertmanager.libsonnet
```diff
@@ -136,6 +137,7 @@ function(params) {
     spec: {
       ports: [
         { name: 'web', targetPort: 'web', port: 9093 },
+        { name: 'reloader-web', port: am._config.reloaderPort, targetPort: am._config.reloaderPort },
```
dgrisonnet, line 140: targetPort should be `reloader-web`

jsonnet/kube-prometheus/components/prometheus.libsonnet
```diff
@@ -98,6 +99,7 @@ function(params) {
     spec: {
       ports: [
                { name: 'web', targetPort: 'web', port: 9090 },
+               { name: 'reloader-web', port: p._config.reloaderPort, targetPort: p._config.reloaderPort },
```
dgrisonnet, line 102: targetPort should be `reloader-web`

dgrisonnet (approved):
Awesome, thank you @PhilipGough.

PR #1365, opened 2021-09-06
Run kubeconform against all  supported versions of Kubernetes

dgrisonnet:
## Description

Instead of running kubeconform on only one version of Kubernetes, it
would be better to run it against the 2 latest versions of Kubernetes
that kube-prometheus supports so that the validation will be in line
with our support policy.

## Type of change

## Changelog entry

```release-note

```


ArthurSens (approved):
LGTM!

But would that mean that when cutting a new release, we would need to update the Makefile too?

If yes, let's just not forget to add that to https://github.com/prometheus-operator/kube-prometheus/pull/1304

dgrisonnet:
I updated #1304 to include a section mentioning that the versions of Kubernetes used by kubeconform need to be updated.

PR #1390, opened 2021-09-21
Add e2e tests for Grafana deployment

ArthurSens:
## Description

Recently, we've been hit by a [change upstream](https://github.com/brancz/kubernetes-grafana/pull/115) that [broke our Grafana deployment](https://github.com/prometheus-operator/kube-prometheus/issues/1372). 

This bug has been fixed by https://github.com/prometheus-operator/kube-prometheus/pull/1373, however, I think that we could add an e2e test to prevent that from happening again.

## Type of change

## Changelog entry

```release-note
NONE
```

### DIFF
--- a/tests/e2e/main_test.go
+++ b/tests/e2e/main_test.go
@@ -277,3 +277,18 @@ func TestFailedRuleEvaluations(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+func TestGrafana(t *testing.T){
+	kClient := promClient.kubeClient
+	
+	err := wait.Poll(30*time.Second, 5*time.Minute, func() (bool, error) {
+		grafanaDeployment, err := kClient.AppsV1().Deployments("monitoring").Get(context.Background(), "grafana", metav1.GetOptions{})
+		if err != nil {
+			t.Fatal(err)
+		}
+		return grafanaDeployment.Status.ReadyReplicas == *grafanaDeployment.Spec.Replicas, nil
+	})
+	if err != nil {
+		t.Fatal(errors.Wrap(err, "Timeout while waiting for deployment ready condition."))
+	}
+}


tests/e2e/main_test.go
```diff
@@ -277,3 +277,42 @@ func TestFailedRuleEvaluations(t *testing.T) {
...
+
+func TestGrafana(t *testing.T){
+	kClient := promClient.kubeClient
+	grafanaScale, err := kClient.AppsV1().Deployments("monitoring").GetScale(context.Background(), "grafana", metav1.GetOptions{})
+	if err != nil {
+		t.Fatal(err)
+	}
+	
+	// Loop for 5 mins
+	for x := 0; x < 10; x++ {
```
dgrisonnet, line 289: For the polling, we should use `wait.Poll` instead of doing it manually. There is an example in the test above if you want to have a look.

tests/e2e/main_test.go
```diff
@@ -277,3 +277,42 @@ func TestFailedRuleEvaluations(t *testing.T) {
...
+		podList, err := kClient.CoreV1().Pods("monitoring").List(context.Background(), metav1.ListOptions{LabelSelector: grafanaScale.Status.Selector})
+		if err != nil {
+			t.Fatal(err)
+		}
+		
+		err = func() error {
+			if len(podList.Items) != int(grafanaScale.Spec.Replicas) {
```
dgrisonnet, line 296: In the podList, you might get pods that are created but not ready so even if there is a problem in the configuration we might not see it. I think it would be better to compare `grafanaScale.Spec.Replicas` to `grafanaScale.Spec.ReadyReplicas`.

dgrisonnet, line 296: In the podList, you might get pods that are created but not ready so even if there is a problem in the configuration we might not see it. I think it would be better to compare `grafanaScale.Spec.Replicas` to `grafanaScale.Spec.ReadyReplicas`.

tests/e2e/main_test.go
```diff
@@ -277,3 +277,42 @@ func TestFailedRuleEvaluations(t *testing.T) {
...
+				return errors.Errorf("Number of pods doesn't match deployment replica count.")
+			}
+		
+			for _, pod := range podList.Items {
+				for _, container := range pod.Status.ContainerStatuses {
+					if !container.Ready {
+						return errors.Errorf("Container %v from pod %v not ready.", container.Name, pod.ObjectMeta.Name)
+					}
+				}
+			}
```
dgrisonnet, line 306: This check will not be needed if we use `ReadyReplicas`.

tests/e2e/main_test.go
```diff
@@ -277,3 +277,42 @@ func TestFailedRuleEvaluations(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+func TestGrafana(t *testing.T){
+	kClient := promClient.kubeClient
+	grafanaScale, err := kClient.AppsV1().Deployments("monitoring").GetScale(context.Background(), "grafana", metav1.GetOptions{})
```
dgrisonnet, line 283: By getting the Deployment instead of the Scale object, you'll be able to get more info about the replicas, such as the number of ReadyReplicas for example.

ArthurSens:
Thanks for the review! The code looks a lot cleaner now :)

dgrisonnet (approved):
Nice improvement @ArthurSens!

PR #1396, opened 2021-09-24
Adjust dropped metrics from cAdvisor

PhilipGough:
This change drops "pod-centric" metrics without the 'container' label.

Previously we dropped Pod centric metrics without a (pod, namespace) label set
however these can be critical for debugging.

## Description

## Type of change

## Changelog entry

```release-note

```

### DIFF
--- a/jsonnet/kube-prometheus/components/k8s-control-plane.libsonnet
+++ b/jsonnet/kube-prometheus/components/k8s-control-plane.libsonnet
@@ -127,9 +127,7 @@ function(params) {
               action: 'drop',
               regex: '(' + std.join('|',
                                     [
-                                      'container_fs_.*',  // add filesystem read/write data (nodes*disks*services*4)
                                       'container_spec_.*',  // everything related to cgroup specification and thus static data (nodes*services*5)
-                                      'container_blkio_device_usage_total',  // useful for containers, but not for system services (nodes*disks*services*operations*2)
                                       'container_file_descriptors',  // file descriptors limits and global numbers are exposed via (nodes*services)
                                       'container_sockets',  // used sockets in cgroup. Usually not important for system services (nodes*services)
                                       'container_threads_max',  // max number of threads in cgroup. Usually for system services it is not limited (nodes*services)
@@ -137,6 +135,15 @@ function(params) {
                                       'container_start_time_seconds',  // container start. Possibly not needed for system services (nodes*services)
                                       'container_last_seen',  // not needed as system services are always running (nodes*services)
                                     ]) + ');;',
+            },
+            {
+              sourceLabels: ['__name__', 'container'],
+              action: 'drop',
+              regex: '(' + std.join('|',
+                                    [  // metrics are available at slice level
+                                      'container_fs_.*',
+                                      'container_blkio_device_usage_total',
+                                    ]) + ');.+',
             },
           ],
         },
--- a/manifests/kubernetes-serviceMonitorKubelet.yaml
+++ b/manifests/kubernetes-serviceMonitorKubelet.yaml
@@ -61,11 +61,16 @@ spec:
       sourceLabels:
       - __name__
     - action: drop
-      regex: (container_fs_.*|container_spec_.*|container_blkio_device_usage_total|container_file_descriptors|container_sockets|container_threads_max|container_threads|container_start_time_seconds|container_last_seen);;
+      regex: (container_spec_.*|container_file_descriptors|container_sockets|container_threads_max|container_threads|container_start_time_seconds|container_last_seen);;
       sourceLabels:
       - __name__
       - pod
       - namespace
+    - action: drop
+      regex: (container_fs_.*|container_blkio_device_usage_total);.+
+      sourceLabels:
+      - __name__
+      - container
     path: /metrics/cadvisor
     port: https-metrics
     relabelings:


simonpasquier:
lgtm

PhilipGough:
Tested in minikube:

Before:
Screenshot 2021-09-24 at 17 31 34


After:
Screenshot 2021-09-24 at 17 34 48

dgrisonnet (approved):

smarterclayton:
lgtm, thanks (in general io behaves like cpu / mem, whereas the others are truly container specific)

mindw:
@PhilipGough, won't these discards metrics related to statefulesets? or pods using host path volumes? are those metrics available elsewhere?
In case they aren't wont it be enough to avoid dropping `container_fs_.*`` metrics all together?
Tx!

PhilipGough:
@mindw my understanding was that `container_fs_.*` metrics would still be available at slice level. However I am open to correction, and if we are losing valuable metrics we can do as you suggested and avoid dropping `container_fs_.*` entirely.

cc @simonpasquier, @smarterclayton

mindw:
@PhilipGough here is an example:

image

dgrisonnet:
Hi @mindw thank you for raising your concerns. To give you more context about what we are trying to achieve here, we had previously discarded some of these metrics as part of an effort to reduce the number of series exposed by the kubelet. So we had identified some container-centric metrics for which pod-level information wasn't useful.
However, in the case of `container_fs_.*` metrics, they make more sense at the pod level since you'll be able to rely on the slice summarization from the cgroups and it should give you as much information as the previous container-centric approach whilst reducing the amount of series exposed by kubelet.
Let me know if it makes sense to you or if you still have some concerns with the new approach.

mindw:
@dgrisonnet thank you for the detailed explanation! My apologies, but I'm not sure how the container level metrics make less sense than pod level - especially with multiple side cars. Also, how does one get those metrics - see:
image

PhilipGough:
@mindw - When I deployed the change to an OpenShift cluster I query a pod in a stateful set and retain the info. So trying to figure out whats different than your setup. Can you provide info on the following:

1. kubernetes version
1. kubelet version
1. the container runtime

Screenshot 2021-09-27 at 12 31 50

mindw:
@PhilipGough 
```
kube_node_info{
container_runtime_version="docker://19.3.15", 
job                                     ="kube-state-metrics", 
kernel_version                   ="5.10.55-flatcar", 
kubelet_version                 ="v1.19.14", 
kubeproxy_version            ="v1.19.14", 
node                                  ="ip-10-0-0-47.eu-west-1.compute.internal", 
os_image                           ="Flatcar Container Linux by Kinvolk 2905.2.1 (Oklo)", 
pod_cidr                            ="10.1.13.0/24", 
provider_id                        ="aws:///eu-west-1a/i-06ed5017ff6f14e1f"
}
```

smarterclayton:
Container level metrics don't completely represent the metrics in a pod - a slice is the sum of all containers plus any work directly handled in the pod slice, so you can't get the total resources accounted to a pod by summing containers.

All that being said, the primary cadvisor metrics are cpu, memory, disk io, and network io (although network is unreliable in most network implementations).  Did we drop container level CPU and memory metrics (I had assumed that was part of the change, but looking at it now it does not look like it)?  If we did not, removing container level io read/write operations/bytes might indeed impact regular users (that's `container_fs_(reads|writes)(_bytes)_total`).

Regarding use of container_name vs container, that looks like a legacydocker/CRI inconsistency.  I'm kind of surprised if that is so, but it may be a known bug (would need to check with cadvisor in forum-node) that they did not change because it would break users.  In general, cadvisor should be returning the same results for different containers because all implementations of the container runtime should be using cgroups in very similar ways.

smarterclayton:
Regarding the pod level metrics being zero, I believe that's a problem in either Docker or the distro configuration with docker.  Pod cgroups are supposed to sum container cgroups, and it's possible this is a docker specific bug.  You'd need to file a separate bug (probably with the distro) if this is an official configuration, or kubernetes if not.  Usually bugs like this are due to cgroup inconsistencies between different linux versions, or a missing configuration delivered to cgroups.

For the short run, including container_fs_* for container level metrics is one option for working around this, but what you are describing is incorrect behavior of your linux distro + container runtime + kubelet.

PhilipGough:
Based on @smarterclayton evaluation, I did some investigation locally (minikube, MacOS), reverting the drop rule to compare the output. I'm not sure if the fact that this is tested on OS X will have any bearing.

Here is what we get on OpenShift, and it gives us what we expected to be available after this change was made:
ocp

___

Checking locally, with docker runtime on minikube, my behaviour matches that reported by @mindw 

mk-docker

___

Changing the driver to cri-o locally produces the following:
mk-crio

Here we notice that we get no 0 value at all, the series without the `container` label disappears entirely.

___

Finally with `containerd` we see the same behaviour as the docker driver

Screenshot 2021-09-27 at 15 48 45

PhilipGough:
@simonpasquier , @dgrisonnet based on the explanation by @smarterclayton I believe this is probably a bug "somewhere" but given that the behaviour is nondeterministic at this point, I think we should revert the drop of `container_fs_.*` until we figure that out. Thoughts?

PR #1402, opened 2021-09-27
Keep 'container_fs_.*' metrics from cAdvisor

PhilipGough:
## Description

Based on the discussion and investigation in https://github.com/prometheus-operator/kube-prometheus/pull/1396, it has become apparent that dropping these particular metrics might become problamatic for users of certain environments who have come to depend on them.

We believed when making the change, that the info provided by these metrics would still be available at a slice level. However in some cases (docker driver?) this appears not to be the case.

As per https://github.com/prometheus-operator/kube-prometheus/pull/1396#issuecomment-927887623 the original change should have been valid but given that it can cause issues, this change reverts the drop.

## Type of change

## Changelog entry

```release-note

```


simonpasquier:
lgtm

dgrisonnet (approved):

PR #1407, opened 2021-09-29
jsonnet/node-exporter: adjust to node-exporter v1.2.0 arg name change

jan--f:
In version [node-exporter v1.2.0](https://github.com/prometheus/node_exporter/releases/tag/v1.2.0)
two argument name changes were introduced. While the old names still
work (with a deprecation warning), lets use the new names.

Signed-off-by: Jan Fajerski <jfajersk@redhat.com>

## Description

Use `--collector.filesystem.mount-points-exclude` instead of deprecated `--collector.filesystem.ignored-mount-points` argument for `node-exporter`.

## Type of change

## Changelog entry

```release-note
- Use `--collector.filesystem.mount-points-exclude` instead of deprecated `--collector.filesystem.ignored-mount-points` argument for `node-exporter`.
```

### DIFF
--- a/jsonnet/kube-prometheus/components/node-exporter.libsonnet
+++ b/jsonnet/kube-prometheus/components/node-exporter.libsonnet
@@ -180,7 +180,7 @@ function(params) {
         '--path.rootfs=/host/root',
         '--no-collector.wifi',
         '--no-collector.hwmon',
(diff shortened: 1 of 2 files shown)


dgrisonnet (approved):
LGTM, thanks @jan--f.

dgrisonnet:
Now that we moved to automated component version update, I guess have been less careful about the changes introduced by each version so it would be great if we could attach the changelogs of the new versions to the PR that does the update. It would be easier to review the changes and prevent this kind of issue from occurring again in the future.

PR #1409, opened 2021-09-29
Drop some of the metrics exposed by prometheus-adapter

dgrisonnet:
## Description

The current implementation of prometheus-adapter exposes a lot of
metrics about the health of its aggregated apiserver. The issue is that
the some of these metrics are not very useful in the context of
prometheus-adapter, and we currently can't avoid exposing them since
they are registered to the Kubernetes global Prometheus registry. Until
this is improved in upstream Kubernetes, we could benefit from dropping
some of the metrics that are not very useful.

Before this change, in a default kube-prometheus installation, we would
have 800+ series for prometheus-adapter against 400+, so we divided the
number of series by two will focusing on the most valuable metrics for
prometheus-adapter.

## Type of change

## Changelog entry

```release-note
Drop some of prometheus-adapter metrics that are inherited from the apiserver code but aren't useful in the context of prometheus-adapter.
```

### DIFF
--- a/jsonnet/kube-prometheus/components/prometheus-adapter.libsonnet
+++ b/jsonnet/kube-prometheus/components/prometheus-adapter.libsonnet
@@ -172,6 +172,21 @@ function(params) {
             insecureSkipVerify: true,
           },
           bearerTokenFile: '/var/run/secrets/kubernetes.io/serviceaccount/token',
+          metricRelabelings: [
+            {
+              sourceLabels: ['__name__'],
+              action: 'drop',
+              regex: '(' + std.join('|',
+                                    [
+                                      'apiserver_client_certificate_.*',  // The only client supposed to connect to the aggregated API is the apiserver so it is not really meaningful to monitor its certificate.
+                                      'apiserver_envelope_.*',  // Prometheus-adapter isn't using envelope for storage.
+                                      'apiserver_flowcontrol_.*',  // Prometheus-adapter isn't using flowcontrol.
+                                      'apiserver_storage_.*',  // Prometheus-adapter isn't using the apiserver storage.
+                                      'apiserver_webhooks_.*',  // Prometeus-adapter doesn't make use of apiserver webhooks.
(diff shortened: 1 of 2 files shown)


dgrisonnet:
@fpetkovski @prashbnair can you please have a look at these changes and let me know if they make sense? Also if you have any other idea of metrics that we could drop let me know.

fpetkovski:
I am not familiar with which metrics are exposed by prometheus-adapter, but would it make sense to make an allowlist instead of a denylist?

dgrisonnet:
I don't think so since it would mean that we would have to think about updating the allowlist whenever we will add new health metrics in prometheus-adapter.

fpetkovski:
Sounds good, then this lgtm

prashbnair:
Do we need the ones related to authentication?

dgrisonnet:
I never used them personally, but I thought that they might be useful if somehow there is an issue with the authentication since the authentication process of aggregated APIs is quite complex and these are the only metrics that we have to investigate, as far as I can tell, we don't have any intel from the apiserver itself since the requests are proxied. That's why I would be reluctant to remove them even though they are responsible for a big part of the series.

Maybe @s-urbaniak can chime in here since he has knowledge on both prometheus-adapter adapter and authentication.

Essentially the metrics that I think are worth keeping are:
- https://github.com/kubernetes/kubernetes/blob/c224ca7674445ec98a2e8876daf53b9846cd3ae1/staging/src/k8s.io/apiserver/pkg/endpoints/filters/metrics.go#L44-L70
- https://github.com/kubernetes/kubernetes/blob/92726bf0f3e2f2c83ca9d0b750cac53a7d74d0c0/staging/src/k8s.io/apiserver/pkg/authentication/authenticatorfactory/metrics.go#L36-L53

But I am in no way an expert on that topic, so I don't really know if these metrics really make sense for an aggregated API.

prashbnair:
/lgtm
lets merge this and we can remove the others if needed.

dgrisonnet:
Thanks everyone for the reviews.

PR #1525, opened 2021-11-25
Source kube-rbac-proxy configs from values

maxbrunet:
## Description

This started off from "I want to easily configure the resources of any kube-rbac-proxy" and end by "we can do the full config for kube-rbac-proxy in values like everything else".

No manifest changes.

## Type of change

## Changelog entry

```release-note
- Source kube-rbac-proxy configs from values
```

### DIFF
--- a/jsonnet/kube-prometheus/components/blackbox-exporter.libsonnet
+++ b/jsonnet/kube-prometheus/components/blackbox-exporter.libsonnet
@@ -11,6 +11,16 @@ local defaults = {
     requests: { cpu: '10m', memory: '20Mi' },
     limits: { cpu: '20m', memory: '40Mi' },
   },
+  kubeRbacProxy:: {
+    name: 'kube-rbac-proxy',
+    upstream: 'http://127.0.0.1:' + defaults.internalPort + '/',
+    resources: defaults.resources,
+    secureListenAddress: ':' + defaults.port,
+    ports: [
+      { name: 'https', containerPort: defaults.port },
+    ],
+    image: defaults.kubeRbacProxyImage,
+  },
   commonLabels:: {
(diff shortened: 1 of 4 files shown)


paulfantom (changes requested):
This is conflicting with https://github.com/prometheus-operator/kube-prometheus/pull/1477 for reasons mentioned in
https://github.com/prometheus-operator/kube-prometheus/issues/1500#issuecomment-968700863.

I would also like to know the use-case of surfacing parameters like `secureListenAddress` or `upstream`.

maxbrunet:
> I would also like to know the use-case of surfacing parameters like secureListenAddress or upstream.

I do not have a use-case at the moment, my idea is to keep it simple and have the whole config in `defaults` and let the user decide whatever they want to use the fields or not. We could also do something like:

```jsonnet
local kubeRbacProxy = krp({
  name: 'kube-rbac-proxy',
  upstream: 'http://127.0.0.1:8080/',
  secureListenAddress: ':8443',
  ports: [
    { name: 'https', containerPort: 8443 },
  ],
  image: defaults.kubeRbacProxyImage,
} po._config.kubeRbacProxy),
```

But it only looks like an extra layer of defaults.

maxbrunet:
> This is conflicting with #1477 for reasons mentioned in [#1500 (comment)](https://github.com/prometheus-operator/kube-prometheus/issues/1500#issuecomment-968700863).

Are referring to this specific sentence? Meaning you have not yet decided on a location to ease configuration of `kube-rbac-proxy`?:

> We could put the kube-rbac-proxy definition into `defaults.containers` list though, but this would still be problematic from UX pov.

What would be the next step to move forward with either of these 2 issues?

FYI something that is important to me is making sure all pods use Guaranteed QoS (limits = requests)

PR #1569, opened 2021-12-30
Update kubeadm integration of kube-prometheus

inductor:
## Description

In short, this PR fixes https://github.com/prometheus-operator/kube-prometheus/issues/1194

It was reported that kubeadm configuration advisory on this repository is very outdated; it uses `kubeadm.k8s.io/v1alpha1` while the upstream is using `kubeadm.k8s.io/v1beta2` and it has no compatibility with the current state.

I addressed this by introducing the latest kubeadm config spec based on https://pkg.go.dev/k8s.io/kubernetes/cmd/kubeadm/app/apis/kubeadm/v1beta2 and each component's args, as well as removing cAdvisor section entirely as it is no longer required due to https://github.com/kubernetes/kubernetes/issues/56523

## Type of change

## Changelog entry

- Enhanced kubeadm integration doc to follow the upstream


paulfantom (approved):
lgtm on green

PR #1586, opened 2022-01-17
build: Bump to build with Go 1.17

PhilipGough:
## Description

I created this because it was blocking the implementation of the security scanning in #1584

However if it is not desirable to merge this we have other options for the linked PR:

1. Choose a tool other than `kubespace`
2.  Run `kubespace` in the action as a container image mounting the manifests dir as a volume.

I choose the version bump and the tooling method since we do similar already but it required a version bump to 1.17 due to  an error `package io/fs is not in GOROOT ` for 1.15

## Type of change

## Changelog entry

```release-note

```

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -3,7 +3,7 @@ on:
   - push
   - pull_request
 env:
-  golang-version: '1.15'
+  golang-version: '1.17'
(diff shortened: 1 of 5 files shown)


paulfantom (approved):
Considering that golang 1.15 is not supported anymore I am more than fine with this change! :100:

PR #1598, opened 2022-01-25
workflows/versions: Enable auto-merge

ArthurSens:
## Description

We've introduced the workflow that updates our dependencies and components version to reduce the maintenance effort of doing so manually.

Looking at all PRs opened by this automation, it doesn't seem that any of them got rejected because it would have broken the stack. With that in mind, I believe we can further automate the workflow by enabling auto-merge, only to the PRs created by this workflow.

Auto-merge needs to be enabled for the repository though, it can be done so in the [repository settings](https://github.com/prometheus-operator/kube-prometheus/settings) if you have access to it (I don't)
image

## Type of change

## Changelog entry

```release-note

```

### DIFF
--- a/.github/workflows/versions.yaml
+++ b/.github/workflows/versions.yaml
@@ -45,6 +45,7 @@ jobs:
         fi
     - name: Create Pull Request
       uses: peter-evans/create-pull-request@v3
+      id: cpr
       with:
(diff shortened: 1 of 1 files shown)


PhilipGough (approved):
lgtm

PR #1647, opened 2022-02-16
Removed CAP_ from node-exporter daemonset

wulfthimm:
## Description

This fixes the `node-exporter` daemonset which had also the `CAP_` prefix which is not expected by Kubernetes according the [this documentation](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/).

## Type of change

## Changelog entry

removed the CAP_ prefix from node-exporter daemonset capabilites

```release-note
- removed the CAP_ prefix from node-exporter daemonset capabilites
```


ArthurSens (approved):
Thanks again for the fix!

Context for others, we're rolling the latest changes to our clusters and discovering some bugs here and there 😅

PR #1675, opened 2022-03-10
Fix problems when enabling eks platform patch

sikachu:
## Description

This PR fixes a problem where users are unable to generate k8s manifests when they enable `platform: 'eks'` in their libsonnet configuration.

An attempt for fixing this has been made in https://github.com/prometheus-operator/kube-prometheus/pull/1286, but it tried to fix all the platforms at once including the documentation. However, in this PR, I would like to focus on just fixing the problem when enabling `platform: 'eks'` part.

Without this patch, user will get an error when they enable `platform: 'eks'` part like this:

```jsonnet
local kp = (import 'kube-prometheus/main.libsonnet') + {
  values+:: {
    common+: {
      namespace: 'monitoring',
      platform: 'eks',
    },
  },
};
```

```
RUNTIME ERROR: Field does not exist: _config
	vendor/kube-prometheus/addons/aws-vpc-cni.libsonnet:80:17-37	
	Field "labels"	
	Field "metadata"	
	Field "resources/kubernetes-prometheusRuleAwsVpcCni"	
	During manifestation
```

To fix the problem, this patch passes `$.values` to make those values available for the platform patches. Then, in `aws-vpc-cni.libsonnet`, I changed those `$.prometheus` to use values from `$.values` or just hard-coded alternate values instead.

I think I will submit follow-up patch to update documentation where necessary to reflect the fact that  you need to set `platform:` inside the `common` object, after this PR is merged.

By the way, this is the first time for me to work with Jsonnet and kube-prometheus codebase, so please feel free to give me feedback if I misunderstood anything. Thank you very much.

## Type of change

## Changelog entry

```release-note
- Fixes a problem when enabling EKS platform patch.
```

### DIFF
--- a/jsonnet/kube-prometheus/addons/aws-vpc-cni.libsonnet
+++ b/jsonnet/kube-prometheus/addons/aws-vpc-cni.libsonnet
@@ -77,9 +77,13 @@
       apiVersion: 'monitoring.coreos.com/v1',
       kind: 'PrometheusRule',
       metadata: {
-        labels: $.prometheus._config.commonLabels + $.prometheus._config.mixin.ruleLabels,
+        labels: {
+          'app.kubernetes.io/name': 'prometheus-vpc-cni-rules',
+          'app.kubernetes.io/component': 'prometheus',
+          'app.kubernetes.io/part-of': 'kube-prometheus',
+        },
         name: 'aws-vpc-cni-rules',
-        namespace: $.prometheus._config.namespace,
+        namespace: $.values.prometheus.namespace,
       },
       spec: {
         groups: [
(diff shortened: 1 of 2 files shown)


jsonnet/kube-prometheus/addons/aws-vpc-cni.libsonnet
```diff
@@ -77,9 +77,13 @@
       apiVersion: 'monitoring.coreos.com/v1',
       kind: 'PrometheusRule',
       metadata: {
-        labels: $.prometheus._config.commonLabels + $.prometheus._config.mixin.ruleLabels,
+        labels: {
+          'app.kubernetes.io/name': 'prometheus-vpc-cni-rules',
+          'app.kubernetes.io/component': 'prometheus',
+          'app.kubernetes.io/part-of': 'kube-prometheus',
+        },
```
sikachu, line 84: For this part, I couldn't find a good replacement for this as we no longer have access to `$.prometheus` in this scope. Please let me know if you have any suggestions on how this can be improved so we don't have to hard-coded these values here.

paulfantom (approved):
lgtm

PhilipGough (approved):

PR #1730, opened 2022-04-14
Access requests to sidecar from thanos-query

polRk:
## Description
Current network polices restrict access to sidecar container from thanos query component. https://github.nmiku.com/thanos-io/kube-thanos/issues/272

## Type of change

## Changelog entry

Allow request from thanos-query to prometheus-operated service.

```release-note
 - Add NetworkPolicy rule allowing requests from thanos-query to prometheus-operated service.
```


ArthurSens:
Hi @polRk, thanks for the fix!

Only one thing is missing, could you run `make generate` to make CI happy? 🙂

ArthurSens:
Maybe we could add this patch only when Thanos is being used as well? Following the rule of least privileged, we don't want to give any access when it's not being used 🤔

polRk:
> Maybe we could add this patch only when Thanos is being used as well? Following the rule of least privileged, we don't want to give any access when it's not being used 🤔

I've been thinking about it. We can enable this rule if the thanos object is specified?

ArthurSens (approved):
Thanks!

PR #1769, opened 2022-05-25
etcd: Add Prometheus alerting and recording rules

centromere:
## Description

The official etcd project provides a [mixin](https://github.com/etcd-io/etcd/tree/main/contrib/mixin) for Prometheus rules and Grafana dashboards. They do not publish auto-generated YAML, so the kube-prometheus-stack project is currently [syncing](https://github.com/prometheus-community/helm-charts/blob/2e86d1618eb5d599576789c1202853dc5bc808c0/charts/kube-prometheus-stack/hack/sync_prometheus_rules.py#L63) against an old ruleset. This ruleset is from the etcd 3.4 documentation, but modern versions of Kubernetes use etcd 3.5+.

## Type of change

## Changelog entry

```release-note
etcd: Add Prometheus alerting and recording rules
```

### DIFF
--- a/jsonnet/kube-prometheus/components/etcd.libsonnet
+++ b/jsonnet/kube-prometheus/components/etcd.libsonnet
@@ -1,0 +1,40 @@
+local defaults = {
+  // Convention: Top-level fields related to CRDs are public, other fields are hidden
+  // If there is no CRD for the component, everything is hidden in defaults.
+  namespace:: error 'must provide namespace',
+  commonLabels:: {
+    'app.kubernetes.io/name': 'kube-prometheus',
+    'app.kubernetes.io/part-of': 'kube-prometheus',
+  },
+  mixin:: {
+    ruleLabels: {},
+    _config: {},
+  },
+};
+
+function(params) {
+  local etcd = self,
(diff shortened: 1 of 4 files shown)


ArthurSens:
Thanks for raising the pull request 🙂

Although I agree that alerting on etcd failures makes sense with this project, the mixin provided by the etcd project is not very compatible with several kubernetes providers 😬.

I just tested deploying the changes from this PR to minikube, kind and gke. All of them don't have the metrics required by the alerts of etcd's mixin

centromere:
If I recall correctly, the metrics come directly from etcd, not the apiserver.

paulfantom (changes requested):
I don't think this should be enabled by default. IMHO there is no point in adding rules when we don't add a ServiceMonitor for etcd. As such I think this should be part of [`static-etcd.libsonnet`](https://github.com/prometheus-operator/kube-prometheus/blob/main/jsonnet/kube-prometheus/addons/static-etcd.libsonnet) addon.

centromere:
WIth https://github.com/prometheus-community/helm-charts/pull/2340 having been merged, this PR is no longer relevant.

PR #1844, opened 2022-08-22
prometheus-adapter: add prefix option to config for container metrics

jan--f:
## Description
This commit adds the options `containerMetricsPrefix`
to the prometheus-adapter config-map generator. By default this option
is the empty string and doesn't change the current behavior. If set
however to e.g. `pa_`, the prometheus-adapter configuration will add
this prefix to all container_ queries in the resource rules.
This enables users of kube-prometheus to define a specialised service
monitor, that only expose the prometheus-adapter related container metrics with a
different configuration, like `honorTimestamps: true` or a tighter
scrape interval.

## Type of change

## Changelog entry

```release-note
- Add `containerMetricsPrefix` to prometheus-adapter.libsonnet to allow customization of prometheus-adapter resource rules 
```

### DIFF
--- a/jsonnet/kube-prometheus/components/prometheus-adapter.libsonnet
+++ b/jsonnet/kube-prometheus/components/prometheus-adapter.libsonnet
@@ -31,6 +31,7 @@ local defaults = {
     nodeExporter: '4m',
     windowsExporter: '4m',
   },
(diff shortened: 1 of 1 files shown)


PhilipGough (commented):
lgtm

PR #1852, opened 2022-08-26
Avoid sorting rules in rule-patching library

dgrisonnet:
## Description

This change improves the rule-sanitizer library to avoid sorting all the
rules. This was originally done to simplify indexing rules with the same
names, but it turns out that integrating this lib can be complex to
review when all the rules of the manifests are reordered. Thus, it is
easier to just leave the file as it was initially sorted.

## Type of change

## Changelog entry

```release-note
Avoid sorting rules in rule-patching library
```

/cc @raptorsun


raptorsun:
Test with a patch and the indexRules() function need some fix. In the returned array from indexRules() the index of every element is 0.
Tested with the following content in `rule-patches.libsonnet`:
``` jsonnet
{
  excludedRuleGroups: [],
  excludedRules: [],
  patchedRules: [
    {
      name: 'kubernetes-storage',
      rules: [
        {
          alert: 'KubePersistentVolumeFillingUp',
          index: 1,
          patches: 'index 1 patch',
        },
      ],
    },
  ],
}
```
In the output file `kubernetes-prometheusRule.yaml` the 2 rules for alert `KubePersistentVolumeFillingUp` has not changed.

dgrisonnet:
It should be good now, I hadn't tested it enough and there was something wrong in the script.

raptorsun:
The new version looks good to me :)

paulfantom (approved):

PR #1891, opened 2022-10-04
Fix alertmanager external config example

godwhoa:
## Description

Problems with the current example:
* Alertmanager can't find `cm/alert-templates` and keeps failing
*  Templates path in external alertmanager config is incorrect
* We are not using the external alertmanager config with templates path included

## Type of change

## Changelog entry

```release-note
- Fix alertmanager external config example
```


philipgough (approved):
lgtm

PR #1900, opened 2022-10-10
Update RAID disk regex

AshtonKem:
Add "md.+" so that raid devices of the form "md123" will match the regex.

## Description

This fixes an issue I have where my raid device is named "md127", which does not match the unmodified regexes. This results in a case where the monitors will not detect if my raid array is degraded or not.

## Type of change

- [X] `BUGFIX` (non-breaking change which fixes an issue)

## Changelog entry

- Modify mdadm regex to include arrays with the naming scheme "mdXXX".

```release-note

```

### DIFF
--- a/manifests/nodeExporter-prometheusRule.yaml
+++ b/manifests/nodeExporter-prometheusRule.yaml
@@ -228,7 +228,7 @@ spec:
         runbook_url: https://runbooks.prometheus-operator.dev/runbooks/node/noderaiddegraded
         summary: RAID Array is degraded
       expr: |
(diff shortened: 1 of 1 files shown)


AshtonKem:
@PhilipGough what are the next steps to getting this reviewed?

PhilipGough:
@AshtonKem your going to need to update the jsonnet and generate the yaml from those changes to pass CI.

AshtonKem:
Ah, I see. I went about this backwards. I'll create a new PR.

PR #1936, opened 2022-11-08
Fixe prometheus-adapter override

JoaoBraveCoding:
Signed-off-by: JoaoBraveCoding <jmarcal@redhat.com>

## Description

Fix override in prometheus-adapter configmap

## Type of change

## Changelog entry

```release-note
fix node resource override in prometheus-adapter configmap
```

### DIFF
--- a/jsonnet/kube-prometheus/components/prometheus-adapter.libsonnet
+++ b/jsonnet/kube-prometheus/components/prometheus-adapter.libsonnet
@@ -88,7 +88,7 @@ local defaults = {
         ||| % { containerMetricsPrefix: $.containerMetricsPrefix },
         resources: {
           overrides: {
-            instance: { resource: 'node' },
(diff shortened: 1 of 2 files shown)


olivierlemasle (changes requested):
No, this is not like in the prometheus-adapter's config you updated:

https://github.com/kubernetes-sigs/prometheus-adapter/blob/268b2a8ec2ab29da82a5f08cc46df35bfd1555d4/deploy/manifests/config-map.yaml#L39-L40

Here, metrics `node_memory_MemTotal_bytes` and `node_memory_MemAvailable_bytes`, which come from the node-exporter, are used. These metrics use label `instance`, not `node`.

So the config is already right :-)

JoaoBraveCoding:
Thank you for the clarification <3 I appreciate it a lot

PR #1968, opened 2022-12-12
fix(deps) update `kube-state-metrics` source repo branch to `main`

jonathan-mothership:
## Description

KSM [updated their repo's primary branch](https://github.com/kubernetes/kube-state-metrics/commit/5d6cb391172960c2aaf98bd729eb55ef0511e0c6) and this repo references `master`. This PR updates that to `main` along with one link in the doc. 

## Type of change

- [ ] `CHANGE` (fix or feature that would cause existing functionality to not work as expected)
- [ ] `FEATURE` (non-breaking change which adds functionality)
- [x] `BUGFIX` (non-breaking change which fixes an issue)
- [ ] `ENHANCEMENT` (non-breaking change which improves existing functionality)
- [x] `NONE` (if none of the other choices apply. Example, tooling, build system, CI, docs, etc.)

## Changelog entry

```release-note
- updated kube-state-metrics source repo branch name
```

PR #1971, opened 2022-12-13
kube-state-metrics switched from master to main

burner-account:
## Description
As kubernetes/kube-state-metrics changed master branch to main branch, jb install currently fails. 

```
jb install <snip>/kube-prometheus@main 
```
Is currently failing with the hint, that kubernetes/kube-state-metrics under the "master" branch creates a nice 404.

## Type of change

I changed version "master" to version "main".

## Changelog entry

Adapted to renamed upstream master branch -> main branch


burner-account:
Does the same as #1968. My bad.

PR #2003, opened 2023-02-03
additional selector for resource queries in Prometheus Adapter.

raptorsun:
## Description

This PR adds 2 optional config fields to add label selectors to the resource queries used by Prometheus Adapter.
When not specified, nothing will change in generated manifests. 
These additional selectors allow users to filter the metrics related to the resource utilization of pods and nodes. 
For example, on some cluster we may want to add a filter `pod!="POD"` when querying metric `container_memory_working_set_bytes`. 
On some clusters, the metric `container_memory_working_set_bytes` can be scraped multiple times for the same pod due to configuration misalignment, we may want to add a filter `job="kubelet"` to make Prometheus Adapter use metrics from a single source.

## Type of change

## Changelog entry

```release-note
Add optional customized selector to resource queries in Prometheus Adapter.
```

### DIFF
--- a/jsonnet/kube-prometheus/components/prometheus-adapter.libsonnet
+++ b/jsonnet/kube-prometheus/components/prometheus-adapter.libsonnet
@@ -34,20 +34,24 @@ local defaults = {
   containerMetricsPrefix:: '',
 
   prometheusURL:: error 'must provide prometheusURL',
+  containerQuerySelector:: '',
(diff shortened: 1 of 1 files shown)


raptorsun:
> lgtm, it's just failing the Doc CI

Thank you :) 
It may be a hiccup from twitter servers when CI is querying that address.

PR #2041, opened 2023-03-13
Improve ArgoCD support

francescsanjuanmrf:
## Description

The `relabelings` block of the `serviceMonitor` configuration for the k8s-control-plane component does not include the `action` field.

The default value for action is `replace`.
- https://github.com/prometheus-operator/prometheus-operator/blob/e2ce57ad6c383a886a7d5fb518a677e1dd6a6115/example/prometheus-operator-crd/monitoring.coreos.com_servicemonitors.yaml#L179-L189
- https://github.com/prometheus-operator/prometheus-operator/blob/e2ce57ad6c383a886a7d5fb518a677e1dd6a6115/example/prometheus-operator-crd/monitoring.coreos.com_servicemonitors.yaml#L360-L373

Having this field undefined makes [ArgoCD](https://argo-cd.readthedocs.io/en/stable/) to be always unsynced. 

The definitions in this repo code lack the `action` field, but when the object is created in Kubernetes, the field is written in the Kubernetes object with the default value (`replace`). Then, ArgoCD compares with the code in the repo and tries to sync again (removing the `action` field). This causes an infinite loop of adding and removing the `action` field.

Setting the `action` field to the default value (`replace`) in the code solves this problem and does not change the behavior of `serviceMonitor`.

## Type of change

## Changelog entry
```
Modifies k8s-control-plane's serviceMonitors to be compatible with [ArgoCD](https://argo-cd.readthedocs.io/en/stable/)
```

### DIFF
--- a/jsonnet/kube-prometheus/components/k8s-control-plane.libsonnet
+++ b/jsonnet/kube-prometheus/components/k8s-control-plane.libsonnet
@@ -105,6 +105,7 @@ function(params) {
           bearerTokenFile: '/var/run/secrets/kubernetes.io/serviceaccount/token',
           metricRelabelings: relabelings,
           relabelings: [{
+            action: 'replace',
             sourceLabels: ['__metrics_path__'],
             targetLabel: 'metrics_path',
           }],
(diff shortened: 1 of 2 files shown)


PhilipGough (approved):
@francescsanjuanmrf change makes sense but you will need to run `make --always-make generate` to generate YAML and pass CI

francescsanjuanmrf:
Just added the modified YAML.

Thanks for your help!

PR #2135, opened 2023-06-07
fix: use names from resources directly

Andor:
## Description

Currently, if you override for any reason Prometheus' ServiceAccount name, the related `subject` name will not change.

The workaround is to override `subject` alongside with overriding ServiceAccount name, which is not ideal.
```
          serviceAccount+: {
            metadata+: {
              name: 'my-serviceaccount-name',
            },
          },
...
          clusterRoleBinding+: {
            subjects: [{
              kind: 'ServiceAccount',
              name: 'my-serviceaccount-name',
              namespace: 'my-namespace',
            }],

```

The same for Role and ClusterRole name in RoleBinding and ClusterRoleBinding accordingly.

This patch fixes this behaviour and you will be able to override

## Type of change

## Changelog entry

```release-note
Use ServiceAccount, Role and ClusterRole names in RoleBinding and ClusterRoleBinding referencing from the according resources instead of generating on the spot.
```

### DIFF
--- a/jsonnet/kube-prometheus/components/prometheus.libsonnet
+++ b/jsonnet/kube-prometheus/components/prometheus.libsonnet
@@ -248,11 +248,11 @@ function(params) {
     roleRef: {
       apiGroup: 'rbac.authorization.k8s.io',
       kind: 'Role',
-      name: p._metadata.name + '-config',
+      name: p.roleConfig.metadata.name,
     },
     subjects: [{
       kind: 'ServiceAccount',
-      name: p._metadata.name,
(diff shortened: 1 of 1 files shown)


Andor:
Apparently tests are failing for non-related reason.

philipgough:
> Apparently tests are failing for non-related reason.

Correct, CI was broken but is back green now. Triggered this again.

Andor:
@philipgough much appreciate mate.
Not it is all green.

philipgough (approved):

PR #2141, opened 2023-06-18
Changing what should be cluster scoped resources from namespace scoped resources

cccsss01:
Deploying this via gitops w/ error checking is causing this deployment to fail.  kubeadm 1.26 (likely other versions as well) fixes this and removes it anyways.

It is failing because it recognizes cluster scoped resources being namespaced.  Kubernetes will auto fix this and deploy it w/o the namespace.  This makes things more compatible/to what I believe correct.

I've simply removed the namespace label from cluster scoped resources.


philipgough (changes requested):
YAML is generated from jsonnet so you will need to make the changes there

Issue #2075, opened 2023-04-13
A CRD is missing when deploying the kube-prometheus-stack with Argo CD


lucj:
When deploying the kube-prometheus-stack with the following definition:
```
apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
  name: kube-prometheus-stack
  namespace: argocd
  finalizers:
    - resources-finalizer.argocd.argoproj.io
spec:
  destination:
    server: https://kubernetes.default.svc
    namespace: monitoring
  project: default
  source:
    repoURL: https://prometheus-community.github.io/helm-charts
    targetRevision: 45.9.1
    chart: kube-prometheus-stack
    helm:
      releaseName: kube-prometheus
      parameters:
      - name: prometheus.prometheusSpec.serviceMonitorSelectorNilUsesHelmValues
        value: "false"
      - name: prometheus.prometheusSpec.serviceMonitorSelector
        value: "{}"
      - name: prometheus.prometheusSpec.serviceMonitorNamespaceSelector
        value: "{}"
  syncPolicy:
    automated:
      prune: true
      selfHeal: true
    syncOptions:
      - CreateNamespace=true
```

I can see the following CRDs are created:
```
$ k get crds | grep coreos
prometheusrules.monitoring.coreos.com       2023-04-13T12:11:36Z
servicemonitors.monitoring.coreos.com       2023-04-13T12:11:36Z
alertmanagerconfigs.monitoring.coreos.com   2023-04-13T12:11:36Z
probes.monitoring.coreos.com                2023-04-13T12:11:36Z
podmonitors.monitoring.coreos.com           2023-04-13T12:11:36Z
alertmanagers.monitoring.coreos.com         2023-04-13T12:11:36Z
thanosrulers.monitoring.coreos.com          2023-04-13T12:11:36Z
```

but I cannot see metrics in the dashboard.
From the logs, the *prometheuses.monitoring.coreos.com* CRD is missing:
```
level=error ts=2023-04-13T12:15:16.836941424Z caller=klog.go:116 component=k8s_client_runtime func=ErrorDepth msg="pkg/mod/k8s.io/client-go@v0.26.1/tools/cache/reflector.go:169: Failed to watch *v1.Prometheus: failed to list *v1.Prometheus: the server could not find the requested resource (get prometheuses.monitoring.coreos.com)"
```

Any idea what could be wrong ?

KGB33:
[...]

PR #2200, opened 2023-08-24
Argo-CD example improvements

roosmaa:
## Description

The current ArgoCD example fell somewhat short for me as I was deploying kube-prometheus on my test cluster. This PR adds the missing pieces that I had to figure out to get it deployed. 

- Using ArgoCD sync-waves to deploy CDRs before everything else
- Work-around big CDRs causing last-applied-configuration annotation too big error (when kubectl apply'ing the CRDs)
- Mention that `application.instanceLabelKey` has to be defined in ArgoCD config map for prometheus target discovery to work properly (by default ArgoCD happily overrides the desired `app.kubernetes.io/instance` value)

## Type of change

## Changelog entry

```release-note
- Update Argo-CD example project
```

### DIFF
--- a/examples/continuous-delivery/argocd/kube-prometheus/argocd-basic.jsonnet
+++ b/examples/continuous-delivery/argocd/kube-prometheus/argocd-basic.jsonnet
@@ -1,14 +1,34 @@
-local kp = (import 'kube-prometheus/main.libsonnet') + {
-  values+:: {
-    common+: {
-      namespace: 'monitoring',
+local kp =
+  (import 'kube-prometheus/main.libsonnet') +
+  // Uncomment the following imports to enable its patches
+  // (import 'kube-prometheus/addons/anti-affinity.libsonnet') +
+  // (import 'kube-prometheus/addons/managed-cluster.libsonnet') +
+  // (import 'kube-prometheus/addons/node-ports.libsonnet') +
+  // (import 'kube-prometheus/addons/static-etcd.libsonnet') +
+  // (import 'kube-prometheus/addons/custom-metrics.libsonnet') +
+  // (import 'kube-prometheus/addons/external-metrics.libsonnet') +
+  // (import 'kube-prometheus/addons/pyrra.libsonnet') +
+  {
+    values+:: {
+      common+: {
+        namespace: 'monitoring',
+      },
     },
-  },
-};
+  };
 
-[kp.kubePrometheus[name] for name in std.objectFields(kp.kubePrometheus)] +
-[kp.prometheusOperator[name] for name in std.objectFields(kp.prometheusOperator)] +
-[kp.nodeExporter[name] for name in std.objectFields(kp.nodeExporter)] +
-[kp.kubeStateMetrics[name] for name in std.objectFields(kp.kubeStateMetrics)] +
-[kp.prometheus[name] for name in std.objectFields(kp.prometheus)] +
-[kp.prometheusAdapter[name] for name in std.objectFields(kp.prometheusAdapter)]
+// Unlike in kube-prometheus/example.jsonnet where a map of file-names to manifests is returned,
+// for ArgoCD we need to return just a regular list with all the manifests.
+local manifests =
+  [ kp.kubePrometheus[name] for name in std.objectFields(kp.kubePrometheus) ] +
+  [ kp.prometheusOperator[name] for name in std.objectFields(kp.prometheusOperator) ] +
+  [ kp.alertmanager[name] for name in std.objectFields(kp.alertmanager) ] +
+  [ kp.blackboxExporter[name] for name in std.objectFields(kp.blackboxExporter) ] +
+  [ kp.grafana[name] for name in std.objectFields(kp.grafana) ] +
+  // [ kp.pyrra[name] for name in std.objectFields(kp.pyrra) ] +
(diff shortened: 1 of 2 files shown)


roosmaa:
Would also address #2075

treenerd:
Hello @roosmaa 

I used kustomize at the past, and now I try to work with jsonnet in case of kube-prometheus.
So please forgive me if I ask basic questions, because I really use jsonnet since yesterday.
I was able to customize and build the manifests like I needed them.
Now I try to get it running with argocd, so I stumpled about your merge request.

I followed the tutorials and build the manifests like described in [customizing kube-prometheus](https://github.com/prometheus-operator/kube-prometheus/blob/v0.11.0/docs/customizing.md).

But when I try to build the [argocd basic example](https://raw.githubusercontent.com/prometheus-operator/kube-prometheus/618a3e554d83a17adf87bf59a566f4748e8e88d6/examples/continuous-delivery/argocd/kube-prometheus/argocd-basic.jsonnet) it fails with folowing error in my case.

```
./build.sh argocd-basic.jsonnet 
+ set -o pipefail
++ pwd
+ PATH=*******
+ rm -rf manifests
+ mkdir -p manifests/setup
+ xargs '-I{}' sh -c 'cat {} | gojsontoyaml > {}.yaml' -- '{}'
+ jsonnet -J vendor -m manifests argocd-basic.jsonnet
RUNTIME ERROR: multi mode: top-level object was a array, should be an object whose keys are filenames and values hold the JSON for that file.
        During manifestation
```

Is this expected in case of jsonnet templates for argocd?
If so, is there a workaround for people like me who would like to create yaml formated manifests?

Thank you
Regards

roosmaa:
> Is this expected in case of jsonnet templates for argocd?
> If so, is there a workaround for people like me who would like to create yaml formated manifests?

@treenerd this was my first jsonnet experience as well. But, I think you have 2 options:

1) If you want to have plain yaml files in your ArgoCD repo, then you use the structure in the `example.jsonnet` and use `build.sh` to get a directory full of kubernetes yaml files (ie effectively what [kube-prometheus/manifests](https://github.com/prometheus-operator/kube-prometheus/tree/main/manifests) folder is, but with your tweaks).

2) You use the built-in jsonnet support that ArgoCD has, in that case you follow directions in examples/continuous-delivery/argocd and you let ArgoCD build what it needs directly from the jsonnet files. (ie what's in this PR.) In this case the build.sh is not needed at all.

I think I saw some mention that 2nd option was the preferred way to do things in some docs around here somewhere, but I'm just a drive-by contributor, so I cannot really say what the suggested way is.

treenerd:
@roosmaa 

Thank you for your answer.
The reason of my reported error was because I configured the wrong path inside my ArgoCD Application for the vendor libraries.
After I red the docs of ArgoCD again, I figured out the root in such a case is the git repo.

[argocd-jsonnet-userguide
](https://argo-cd.readthedocs.io/en/stable/user-guide/jsonnet/)
```
It is also possible to add a shared library (e.g. vendor folder) relative to the repository root.
```

Never the less I had other issues, so I decided to build manifests and kustomize overlays with jsonnet.
So I use the kustomization.yaml for the rollouts within my ApplicationSets..

Thank you for your fast response and help.

philipgough:
@roosmaa just a heads up this is failing ci if you want to address it :)

roosmaa:
@philipgough on my todo list :) haven't managed to open the workstation in a few evenings.

examples/continuous-delivery/argocd/kube-prometheus/argocd-basic.jsonnet
```diff
@@ -1,14 +1,66 @@
...
+  [kp.blackboxExporter[name] for name in std.objectFields(kp.blackboxExporter)] +
+  [kp.grafana[name] for name in std.objectFields(kp.grafana)] +
+  // [ kp.pyrra[name] for name in std.objectFields(kp.pyrra)] +
+  [kp.kubeStateMetrics[name] for name in std.objectFields(kp.kubeStateMetrics)] +
+  [kp.kubernetesControlPlane[name] for name in std.objectFields(kp.kubernetesControlPlane)] +
+  [kp.nodeExporter[name] for name in std.objectFields(kp.nodeExporter)] +
+  [kp.prometheus[name] for name in std.objectFields(kp.prometheus)] +
+  [kp.prometheusAdapter[name] for name in std.objectFields(kp.prometheusAdapter)];
+
+local argoAnnotations(manifest) =
```
roosmaa, line 38: @philipgough would it make sense to upgrade this approach to an addon, ie `kube-prometheus/addons/argo-cd.libsonnet`?

Asking because if it's an addon, then it would be somewhat more of a maintenance burden for kube-prometheus and I have no idea if the project wants to take that on.

philipgough: > then it would be somewhat more of a maintenance burden for kube-prometheus
it would be

> no idea if the project wants to take that on
not particularly right now, because I am short on time and the project is short on maintainers as it is 😅

roosmaa: Fine by me 👍 (less PRs for me to do haha)

philipgough (approved):

PR #2238, opened 2023-10-08
add monitor and rules resources to user-facing roles

jouve:
## Description

in cluster with separation between (cluster) admin and (namespaced) users, it allows the namespaced users to create monitor and rules in their namespaces according to the default k8s model of user-facing roles.

defined as an addon to have the feature disabled by default and simple to enable.

ref: https://kubernetes.io/docs/reference/access-authn-authz/rbac/#user-facing-roles

## Type of change

## Changelog entry

```release-note
- add monitor and rules resources to user-facing roles (addon)
```


philipgough (approved):
/lgtm

PR #2263, opened 2023-11-09
fix: label mismatch for AlertmanagerFailedToSendAlerts

cr7258:
## Description

Since [Alertmanager 0.26.0 version](https://github.com/prometheus/alertmanager/blob/ce6efba023b0397cb522d64e910684e48d12455f/CHANGELOG.md#0260--2023-08-23), a new label `reason` was added in the `alertmanager_notifications_failed_total` metric to indicate the type of error of the alert delivery. 

As a result, the original alert rules are broken because labels are mismatched between `alertmanager_notifications_failed_total` and `alertmanager_notifications_total metrics`.
Prometheus requires samples with exactly the same labels to get matched together when performing calculations. [docs](https://prometheus.io/docs/prometheus/latest/querying/operators/#one-to-one-vector-matches)

Use the `ignoring` vector matching keyword to ignore the new `reason` label to allow for matching between series with different labels.

## Type of change

## Changelog entry

```release-note
Fix label mismatch for AlertmanagerFailedToSendAlerts
```

### DIFF
--- a/docs/access-ui.md
+++ b/docs/access-ui.md
@@ -1,8 +1,19 @@
-# Access UIs
+---
+weight: 300
+toc: true
+title: Access Dashboards
+menu:
+    docs:
+        parent: kube
+images: []
+draft: false
+---
 
-Prometheus, Grafana, and Alertmanager dashboards can be accessed quickly using `kubectl port-forward` after running the quickstart via the commands below. Kubernetes 1.10 or later is required.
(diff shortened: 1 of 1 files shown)


simonpasquier:
Thanks for spotting it! The fix needs to happen in prometheus/alertmanager first though.

cr7258:
@simonpasquier The [PR](https://github.com/prometheus/alertmanager/pull/3599) has been merged in prometheus/alertmanager repo. Can we proceed with this PR now?

philipgough:
I think we will need a release in AM first.

PR #2294, opened 2023-12-01
script/generate-versions: fix versions sort

dgrisonnet:
## Description

I was wondering why kube-state-metrics wasn't updated from v2.9.2 to v2.10.z despite these new releases out for a couple a few months, and it seems that the script was buggy.

Instead of using `-n` option we should use `-V` which is meant for sorting versions.

```sh
➜  ~ printf '%s\n' "9" "10" | sort -V -r 
10
9
➜  ~ printf '%s\n' "2.9.2" "2.10.1" | sort -V -r
2.10.1
2.9.2
```

This new version also fixes the problem reported in https://github.com/prometheus-operator/kube-prometheus/pull/2261.

## Type of change

## Changelog entry

```release-note

```

### DIFF
--- a/scripts/generate-versions.sh
+++ b/scripts/generate-versions.sh
@@ -34,7 +34,7 @@ get_version() {
   fi
 
   # Use higher version from new version and current version
-  v=$(printf '%s\n' "$v" "$cv" | sort -n -r | head -n1)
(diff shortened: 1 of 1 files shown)


dgrisonnet:
cc @philipgough 

If this change didn't make it to the next trigger of the script, would you mind to retrigger it manually? :bow:

philipgough (approved):
lgtm

Issue #2300, opened 2023-12-07
 podmonitor failed to list *v1.Pod: pods is forbidden


shaoxiahuiwugong:
podmonitor system:serviceaccount:monitoring:prometheus-k8s is forbidden

 list *v1.Endpoints: endpoints is forbidden: User \"system:serviceaccount:monitoring:prometheus-k8s\" cannot list resource \"endpoints\" in API group \"\" in the namespace \"trec\""
ts=2023-12-06T02:16:28.779Z caller=klog.go:108 level=warn component=k8s_client_runtime func=Warningf msg="pkg/mod/k8s.io/client-go@v0.24.0/tools/cache/reflector.go:167: failed to list *v1.Endpoints: endpoints is forbidden: User \"system:serviceaccount:monitoring:prometheus-k8s\" cannot list resource \"endpoints\" in API group \"\" in the namespace \"tb\""
ts=2023-12-06T02:16:28.779Z caller=klog.go:116 level=error component=k8s_client_runtime func=ErrorDepth msg="pkg/mod/k8s.io/client-go@v0.24.0/tools/cache/reflector.go:167: Failed to watch *v1.Endpoints: failed to list *v1.Endpoints: endpoints is forbidden: User \"system:serviceaccount:monitoring:prometheus-k8s\" cannot list resource \"endpoints\" in API group \"\" in the namespace \"tb\""
ts=2023-12-06T02:16:31.062Z caller=klog.go:108 level=warn component=k8s_client_runtime func=Warningf msg="pkg/mod/k8s.io/client-go@v0.24.0/tools/cache/reflector.go:167: failed to list *v1.Pod: pods is forbidden: User \"system:serviceaccount:monitoring:prometheus-k8s\" cannot list resource \"pods\" in API group \"\" in the namespace \"tap\""

PR #2301, opened 2023-12-07
fix service account is forbidden

shaoxiahuiwugong:
fix podmonitor failed to list *v1.Pod: pods is forbidden 
#2300


philipgough:
This change needs to be made in jsonnet and the yaml generated off the back of that. Also this change looks to wide open in terms of RBAC permissions to solve the reported issue. Should only need get list and watch on Pods and Endpoints.

PR #2450, opened 2024-06-27
Multi arch win exporter

henrik-m:
Make windows-exporter compatible with multiple Windows versions

### DIFF
--- a/jsonnet/kube-prometheus/addons/windows-hostprocess.libsonnet
+++ b/jsonnet/kube-prometheus/addons/windows-hostprocess.libsonnet
@@ -67,7 +67,7 @@ local windowsExporter = function(params) {
           initContainers: [
             {
               name: 'configure-firewall',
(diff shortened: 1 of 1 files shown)


philipgough:
@henrik-m - This looks fine to me but I am not a Windows user. Care to add some context or clarify consequences? Thanks

henrik-m:
@philipgough These changes allow the Windows Exporter to run not only on Windows Server 2019, but also on 2022+. The init container was changed to an image with a multi-arch manifest.
The toleration was added to allow the exporter to run in hybrid clusters where the Windows nodes are tainted with os=windows so that linux kubernetes workloads do not get scheduled on them.
And the argument change, iirc, was just a straight fix for more recent versions of the windows exporter image, because the `collector.textfile.directory` option was [deprecated](https://github.com/prometheus-community/windows_exporter/blob/master/docs/collector.textfile.md).

philipgough (approved):
/lgtm

PR #2467, opened 2024-07-29
Improve watchdog alert

a-Tell:
## Description
This expression fires the Watchdog alert also if the TSDB is up to date and therefore checks the functionality of the full stack.

With `vector(1)`, only alertmanager needs to be functional to fire. So In case of a full TSDB storage, the Watchdog still fires and the lack of new metrics goes unnoticed.

## Type of change

## Changelog entry

```release-note
Change Watchdog alert to check if TSDB is up-to-date
```


a-Tell:
Hm, I honestly don't know why the Ubuntu-Pipeline exited with 1. Just the Rule-Expression changed. Maybe it's an issue with the pipeline itself?

paulfantom:
> Hm, I honestly don't know why the Ubuntu-Pipeline exited with 1

You need to run `make generate` and commit generated code changes too.

---

The initial purpose of the Watchdog alert was to ensure general ability to send alerts is operational. For this reason, `vector(1)`, which was not touching prometheus tsdb, seemed like a good choice. This is somewhat confirmed by lookingat the alert summary which states nothing about the prometheus itself - `An alert that should always be firing to certify that Alertmanager is working properly.`. If runbook states something else, then I do believe it is a bug in the runbook itself.

As for testing if prometheus can generate alerts, I think this should be a part of prometheus mixin and if I am correct, one of `PrometheusMissingRuleEvaluations`, `PrometheusRuleFailures`, `PrometheusNotIngestingSamples` would fire in that case. This in turn gives a full picture - simple Watchdog for alert sending and additional alerts for rule evaluation.

a-Tell:
Hey @paulfantom, thanks alot for the feedback!

In my years of experiencing "dead" Prometheus'es, I never saw one of the Alerts you mentioned firing. I'm gonna check on them.

And I agree with you, keeping the Watchdog-Alert as it is (and only Reporting on a down Alertmanager) and having another Alert which tells that there is an Issue with Rule-Evaluation would be a nice solution.

Thanks again!

PR #2485, opened 2024-08-12
Allow specifying which node exporter collectors to exclude

Aeyk:
## Description

Allow specifying which collectors to exclude from node-exporter. 

## Type of change

## Changelog entry

Parameter for customizing which collectors to disable in node-exporter.

```release-note
- Add jsonnet parameter to allow customizing which node-exporter collectors to disable. 
```

### DIFF
--- a/jsonnet/kube-prometheus/components/node-exporter.libsonnet
+++ b/jsonnet/kube-prometheus/components/node-exporter.libsonnet
@@ -21,6 +21,7 @@ local defaults = {
   },
   listenAddress:: '127.0.0.1',
   filesystemMountPointsExclude:: '^/(dev|proc|sys|run/k3s/containerd/.+|var/lib/docker/.+|var/lib/kubelet/pods/.+)($|/)',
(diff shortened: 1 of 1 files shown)


jsonnet/kube-prometheus/components/node-exporter.libsonnet
```diff
@@ -21,6 +21,7 @@ local defaults = {
   },
   listenAddress:: '127.0.0.1',
   filesystemMountPointsExclude:: '^/(dev|proc|sys|run/k3s/containerd/.+|var/lib/docker/.+|var/lib/kubelet/pods/.+)($|/)',
+  exportersExclude:: ['wifi', 'btrfs', 'hwmon'],
```
slashpai, line 24: IIUC by default we should not exclude these, you can have the exclude variable and while importing this jsonnet you should define the variable to include collectors to exclude so that default setting doesn't change

Aeyk: They were already excluded, I just exposed them as a variable.

PR #2495, opened 2024-08-26
Update grafana-dashboardDatasources.yaml 

arohyadav:
## Description

This pull request addresses an issue where the Prometheus data source does not work out of the box when Grafana is configured with the URL http://prometheus-k8s.monitoring.svc:9090. The data source works correctly when using the URL http://prometheus-operated.monitoring.svc:9090. This change ensures that the correct service URL is used for the Prometheus data source in Grafana, improving the out-of-the-box experience.

## Type of change

### DIFF
--- a/manifests/grafana-dashboardDatasources.yaml
+++ b/manifests/grafana-dashboardDatasources.yaml
@@ -19,7 +19,7 @@ stringData:
                 "name": "prometheus",
                 "orgId": 1,
                 "type": "prometheus",
-                "url": "http://prometheus-k8s.monitoring.svc:9090",
+                "url": "http://prometheus-operated.monitoring.svc:9090",
                 "version": 1
             }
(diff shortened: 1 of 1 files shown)


manifests/grafana-dashboardDatasources.yaml
```diff
@@ -19,7 +19,7 @@ stringData:
                 "name": "prometheus",
                 "orgId": 1,
                 "type": "prometheus",
-                "url": "http://prometheus-k8s.monitoring.svc:9090",
+                "url": "http://prometheus-operated.monitoring.svc:9090",
```
slashpai, line 22: The service name is based on name defined for prometheus service in manifest https://github.com/prometheus-operator/kube-prometheus/blob/main/manifests/prometheus-service.yaml#L10

PR #2505, opened 2024-09-08
fix: allow opting-into upstream probes

rexagod:
## Description

Allow users to opt-into upstream probe definitions.

## Type of change

## Changelog entry

```release-note
Users can opt-into the upstreamProbesOptIn (boolean, false by default) for KSM to sync with upstream's health probe definitions.
```

### DIFF
--- a/jsonnet/kube-prometheus/components/kube-rbac-proxy.libsonnet
+++ b/jsonnet/kube-prometheus/components/kube-rbac-proxy.libsonnet
@@ -38,6 +38,13 @@ local defaults = {
     'TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305',
     'TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305',
   ],
+  // Corresponds to KRP's --ignore-paths flag.
+  // Some components (for e.g., KSM) may utilize the flag to allow for communication with external parties in scenarios
+  // where the originating request(s) cannot be modified to the proxy's expectations, and thus, are passed through, as
+  // is, to certain endpoints that they target, without the proxy's intervention. The kubelet, in KSM's case, can thus
+  // query health probe endpoints without being blocked by KRP, thus allowing for http-based probes over exec-based
+  // ones.
+  ignorePaths:: [],
 };
 
 
@@ -50,10 +57,11 @@ function(params) {
   name: krp._config.name,
   image: krp._config.image,
   args: [
-    '--secure-listen-address=' + krp._config.secureListenAddress,
-    '--tls-cipher-suites=' + std.join(',', krp._config.tlsCipherSuites),
-    '--upstream=' + krp._config.upstream,
-  ],
+          '--secure-listen-address=' + krp._config.secureListenAddress,
+          '--tls-cipher-suites=' + std.join(',', krp._config.tlsCipherSuites),
+          '--upstream=' + krp._config.upstream,
+        ]  // Optionals.
+        + if std.length(krp._config.ignorePaths) > 0 then ['--ignore-paths=' + std.join(',', krp._config.ignorePaths)] else defaults.ignorePaths,
   resources: krp._config.resources,
   ports: krp._config.ports,
   securityContext: {
--- a/jsonnet/kube-prometheus/components/kube-state-metrics.libsonnet
+++ b/jsonnet/kube-prometheus/components/kube-state-metrics.libsonnet
@@ -15,12 +15,18 @@ local defaults = {
   },
 
   kubeRbacProxyMain:: {
+    ports: [
+      { name: 'http-metrics', containerPort: 8443 },
+    ],
     resources+: {
       limits+: { cpu: '40m' },
       requests+: { cpu: '20m' },
     },
   },
   kubeRbacProxySelf:: {
+    ports: [
+      { name: 'telemetry', containerPort: 9443 },
+    ],
     resources+: {
       limits+: { cpu: '20m' },
       requests+: { cpu: '10m' },
@@ -46,6 +52,8 @@ local defaults = {
       runbookURLPattern: 'https://runbooks.prometheus-operator.dev/runbooks/kube-state-metrics/%s',
     },
   },
+  // `enableProbes` allows users to opt-into upstream definitions for health probes.
+  enableProbes:: false,
 };
 
 function(params) (import 'github.com/kubernetes/kube-state-metrics/jsonnet/kube-state-metrics/kube-state-metrics.libsonnet') {
@@ -91,14 +99,14 @@ function(params) (import 'github.com/kubernetes/kube-state-metrics/jsonnet/kube-
     spec+: {
       ports: [
         {
-          name: 'https-main',
-          port: 8443,
-          targetPort: 'https-main',
+          name: defaults.kubeRbacProxyMain.ports[0].name,
+          port: defaults.kubeRbacProxyMain.ports[0].containerPort,
+          targetPort: defaults.kubeRbacProxyMain.ports[0].name,
         },
         {
-          name: 'https-self',
-          port: 9443,
-          targetPort: 'https-self',
+          name: defaults.kubeRbacProxySelf.ports[0].name,
+          port: defaults.kubeRbacProxySelf.ports[0].containerPort,
+          targetPort: defaults.kubeRbacProxySelf.ports[0].name,
         },
       ],
     },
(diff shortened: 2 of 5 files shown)


rexagod:
Tested downstream (CMO), on enabling `upstreamProbesOptIn`.

```diff
diff --git a/assets/kube-state-metrics/deployment.yaml b/assets/kube-state-metrics/deployment.yaml
index ec651432e..e73091269 100644
--- a/assets/kube-state-metrics/deployment.yaml
+++ b/assets/kube-state-metrics/deployment.yaml
@@ -57,7 +57,19 @@ spec:
           ^kube_pod_completion_time$,
           ^kube_pod_status_scheduled$
         image: registry.k8s.io/kube-state-metrics/kube-state-metrics:v2.13.0
+        livenessProbe:
+          httpGet:
+            path: /livez
+            port: 8080
+          initialDelaySeconds: 5
+          timeoutSeconds: 5
         name: kube-state-metrics
+        readinessProbe:
+          httpGet:
+            path: /livez
+            port: 8080
+          initialDelaySeconds: 5
+          timeoutSeconds: 5
         resources:
           requests:
             cpu: 2m
diff --git a/jsonnet/main.jsonnet b/jsonnet/main.jsonnet
index 7539fb456..ffabfb4f0 100644
--- a/jsonnet/main.jsonnet
+++ b/jsonnet/main.jsonnet
@@ -216,6 +216,7 @@ local inCluster =
         kubeRbacProxyImage: $.values.common.images.kubeRbacProxy,
         commonLabels+: $.values.common.commonLabels,
         mixin+: { ruleLabels: $.values.common.ruleLabels },
+        upstreamProbesOptIn: true,
       },
       nodeExporter: {
         namespace: $.values.common.namespace,
@@ -439,7 +440,7 @@ local inCluster =
     openshiftStateMetrics: openshiftStateMetrics($.values.openshiftStateMetrics),
   } +
   (import './utils/anti-affinity.libsonnet') +
-  (import 'github.com/prometheus-operator/kube-prometheus/jsonnet/kube-prometheus/addons/ksm-lite.libsonnet') +
+  (import 'github.com/rexagod/kube-prometheus/jsonnet/kube-prometheus/addons/ksm-lite.libsonnet') +
   (import './utils/ibm-cloud-managed-profile.libsonnet') +
   (import './components/metrics-server-audit.libsonnet') +
   {};  // Including empty object to simplify adding and removing imports during development

```

philipgough (approved):
/lgtm

jsonnet/kube-prometheus/components/kube-state-metrics.libsonnet
```diff
@@ -162,13 +163,17 @@ function(params) (import 'github.com/kubernetes/kube-state-metrics/jsonnet/kube-
...
               runAsGroup: 65534,
             },
             args: ['--host=127.0.0.1', '--port=8081', '--telemetry-host=127.0.0.1', '--telemetry-port=8082'],
             resources: ksm._config.resources,
+          } + if ksm._config.upstreamProbesOptIn then {
+            livenessProbe: super.livenessProbe,
+            readinessProbe: super.livenessProbe,
+          } else {
+            livenessProbe:: null,
+            readinessProbe:: null,
```
simonpasquier, line 176:
```suggestion
          } + if !ksm._config.upstreamProbesOptIn then {
            livenessProbe:: null,
            readinessProbe:: null,
          } else {
```

simonpasquier (commented):
Shouldn't we also adjust the kube-rbac-proxy args to include the proper URLs in `--ignore-paths`?

jsonnet/kube-prometheus/components/kube-rbac-proxy.libsonnet
simonpasquier, line 1: It might be easier to add a `ignorePaths` field to the config and default to `[]`?

rexagod: Ah sorry I saw this after the push, ACK.

rexagod:
- [ ] Add a comment describing the way probes are defined and expected to work.

jsonnet/kube-prometheus/components/kube-state-metrics.libsonnet
```diff
@@ -122,6 +124,7 @@ function(params) (import 'github.com/kubernetes/kube-state-metrics/jsonnet/kube-
       { name: 'https-self', containerPort: 9443 },
     ],
     image: ksm._config.kubeRbacProxyImage,
+    ignorePaths: if ksm._config.upstreamProbesOptIn then ['/readyz'] else super.ignorePaths,
```
simonpasquier, line 127:
```suggestion
    ignorePaths: if ksm._config.upstreamProbesOptIn then ['/readyz'] else super.ignorePaths,
```
```suggestion
    // When enabling probes, kube-rbac-proxy needs to always allow the /readyz endpoint.
    ignorePaths: if ksm._config.upstreamProbesOptIn then ['/readyz'] else [],
```

jsonnet/kube-prometheus/components/kube-state-metrics.libsonnet
```diff
@@ -46,6 +46,7 @@ local defaults = {
       runbookURLPattern: 'https://runbooks.prometheus-operator.dev/runbooks/kube-state-metrics/%s',
     },
   },
+  upstreamProbesOptIn:: false,
```
simonpasquier, line 49: a comment explaining the purpose + implementation would be useful here.
```suggestion
  enableProbes:: false,
```

rexagod:
@simonpasquier Pushed all the suggested changes but `s/defaults.ignorePaths/[]` and `s/super.ignorePaths/[]` since will always be `[]`, in addition to the current implementation allowing for a single source of truth for the default value?

simonpasquier:
I generated the manifest with `enableProbes: true` but Kubernetes isn't happy with it:

````
$ kc events
LAST SEEN                TYPE      REASON      OBJECT                                    MESSAGE
6m20s (x2355 over 21h)   Warning   Unhealthy   Pod/kube-state-metrics-68fccdc5f7-9wzdn   Liveness probe errored: strconv.Atoi: parsing "http-metrics": invalid syntax
80s (x2703 over 21h)     Warning   Unhealthy   Pod/kube-state-metrics-68fccdc5f7-9wzdn   Readiness probe errored: strconv.Atoi: parsing "telemetry": invalid syntax
````

**Comment by rexagod** [comment]
The probes seem to work now, my guess is that due to the names being hardcoded (for KRP), they had to be updated, the lack of which made the controller-manager parse the port strings as numbers since they weren't defined beforehand. It looks like the workflows need to be jumpstarted, though.

**Comment by rexagod** [comment]
Tested on a cluster, seems to work now. I've made KRP's ports the single source of truth for KSM's probes and ports for convenience.

**Comment by rexagod** [comment]
Needs an approval (for workflows).

**Comment by simonpasquier** [comment]
https://github.com/prometheus-operator/kube-prometheus/actions/runs/11370566998/job/31653537858?pr=2505 fails on kube-state-metrics check for metrics.

**Comment by rexagod** [comment]
[Fixed](https://github.com/rexagod/kube-prometheus/actions/runs/11442422799), needs workflow approval here

**Comment by rexagod** [comment]
Re-ping, @simonpasquier for another look here. 🙁

**Review by simonpasquier — COMMENTED** [review]
I still see warnings when applying the manifests with `enableProbes: true`.

````
Warning: spec.template.spec.containers[1].ports[0]: duplicate port definition with spec.template.spec.containers[0].ports[0]
Warning: spec.template.spec.containers[2].ports[0]: duplicate port definition with spec.template.spec.containers[0].ports[1]
```

Also it still feels strange that the probes being defined on the ksm container while the kubelet queries the kube-rbac-proxy containers. I understand that for the liveness probe, it's required so we restart the ksm container if it gets stuck but could we at least define the readiness probe in the ksm container?

rexagod:
Since port definitions use the `containerPort` as keys, I'm not sure if the warnings can be addressed. Since the APIServer sees this as an anti-pattern, a potential fix may come from KRP. I'll see if we can do something there.

rexagod:
Since port-definitions use the port numbers as keys, I believe there's no workaround to define the same ones on KSM, without server-side apply warnings. I'll open a ticket/PR upstream to see if we cannot address this there.

rexagod:
Actually, since Kubelet looks at KSM's ports and pings them on the proxy (which then goes through successfully if the port definitions are duplicated) may suggest that we'd need to bring this up to the upstream k/k community to address this particular use-case.

rexagod:
It seems like we've hit a dead end here, since duplicate port definitions are not allowed within the same pod.

To think out loud, one way to accomplish (and simplify) this would be if KRP were to move out of the picture, and we'd move to KSM's [`WithAuthenticationAndAuthorization`](https://github.com/kubernetes/kube-state-metrics/pull/2686) implementation instead.

I'll close this for the time being, but feel free to reopen if there's anything I can help with here.

PR #2575, opened 2024-12-23
Fix broken rules with Prometheus v3

leonnicolas:
With Prometheus v3 [histograms are normalized](https://prometheus.io/docs/prometheus/latest/migration/#le-and-quantile-label-values) and this breaks rules that
select on the `le` label.
This PR overwrites the label selectors that
[mixin](https://github.com/kubernetes-monitoring) usese.

Before:
```
{le="1"}
```
Now with Prometheus v3:
```
{le="1.0"}
```

This will break set ups with prometheus v2.
However, I think it is still an improvement by default Prometheus v3 is
used and I would expect that the generated rules are workind with the
default Prometheus version as well.

Fixes https://github.com/prometheus-operator/kube-prometheus/issues/2573

Signed-off-by: leonnicolas <leonloechner@gmx.de>


leonnicolas:
the rules were updated upstream in https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/1008. So maybe we can just bump a version and close this PR

leonnicolas:
fixed by #2581

PR #2627, opened 2025-03-17
Fix port allocation for node-exporter

steve-wolter-ai:
## Description

The daemonset for the node exporter is statically invalid. It assigns port 9100 both to the node-exporter container and to the kube-rbac-proxy container.

This results in the following behavior at runtime:

  1. kube-rbac-proxy and node-exporter race for port 9100.
  2. If kube-rbac-proxy wins, the system functions normally. However, kube-rbac-proxy is disfunctional and does NOT actually check authorization.
  3. If kube-rbac-proxy wins, it fails to locate its upstream and crashes. The race restarts.

This change is somewhat risky as we've been (so far) rolling out node-exporter with no kube-rbac-proxy, and would now unmasks any bugs in the kube-rbac-proxy integration.

## Type of change

## Changelog entry

Activate kube-rbac-proxy for node-exporter.

```release-note

```

### DIFF
--- a/jsonnet/kube-prometheus/components/node-exporter.libsonnet
+++ b/jsonnet/kube-prometheus/components/node-exporter.libsonnet
@@ -26,6 +26,7 @@ local defaults = {
   // thus [a-z0-9] regex below
   ignoredNetworkDevices:: '^(veth.*|[a-f0-9]{15})$',
   port:: 9100,
+  internal_port:: 9101,
   commonLabels:: {
     'app.kubernetes.io/name': defaults.name,
     'app.kubernetes.io/version': defaults.version,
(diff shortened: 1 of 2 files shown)


cypres:
You meant
> If node-exporter wins, the system functions normally. However, kube-rbac-proxy is disfunctional and does NOT actually check authorization.

Right

philipgough (approved):
/lgtm thanks

simonpasquier:
Resurrecting this PR because it might have broken something for us and I have questions around this statement:

> The daemonset for the node exporter is statically invalid. It assigns port 9100 both to the node-exporter container and to the kube-rbac-proxy container.

The 2 containers bind to different IP addresses: node_exporter uses 127.0.0.1 while krp uses the pod's IP address. @steve-wolter-ai was there something in your setup that didn't work the same?

PR #2655, opened 2025-05-15
Update go 1.24

simonpasquier:
## Type of change

## Changelog entry

```release-note

```

### DIFF
--- a/scripts/go.mod
+++ b/scripts/go.mod
@@ -1,7 +1,8 @@
 module _ // go.mod created for tooling dependencies
 
-go 1.22
-toolchain go1.24.1
+go 1.23.7
+
+toolchain go1.24.3
 
 require (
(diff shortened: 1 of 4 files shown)


philipgough (approved):
/lgtm

scripts/go.mod
```diff
@@ -1,7 +1,8 @@
 module _ // go.mod created for tooling dependencies
 
-go 1.22
-toolchain go1.24.1
+go 1.23.7
```
slashpai, line 3: should be `go 1.24.2`?

simonpasquier: https://go.dev/doc/modules/gomod-ref#go

`The go line must be greater than or equal to the go line of all dependencies.`

IIUC the Go toolchain selected 1.23 because it is the min required version from all deps.

PR #2659, opened 2025-05-16
chore: bump tested kubernetes versions

simonpasquier:
## Description

This change updates the end-to-end tests to use the 3 latest Kubernetes versions:
* 1.31
* 1.32
* 1.33

## Type of change

## Changelog entry

```release-note

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,13 +41,13 @@ no effect, but is still deployed.
 
 The following Kubernetes versions are supported and work as we test against these versions in their respective branches. But note that other versions might work!
 
-| kube-prometheus stack                                                                      | Kubernetes 1.23 | Kubernetes 1.24 | Kubernetes 1.25 | Kubernetes 1.26 | Kubernetes 1.27 | Kubernetes 1.28 | Kubernetes 1.29 | Kubernetes 1.30 | Kubernetes 1.31 |
-|--------------------------------------------------------------------------------------------|-----------------|-----------------|-----------------|-----------------|-----------------|-----------------|-----------------|-----------------|-----------------|
-| [`release-0.11`](https://github.com/prometheus-operator/kube-prometheus/tree/release-0.11) | ✔               | ✔               | ✗               | x               | x               | x               | x               | x               | x               |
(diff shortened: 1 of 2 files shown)


slashpai (approved):

README.md
```diff
@@ -41,13 +41,13 @@ no effect, but is still deployed.
...
+| [`release-0.12`](https://github.com/prometheus-operator/kube-prometheus/tree/release-0.12) | ✗               | ✔               | ✔               | x               | x               | x               | x               | x               | x               | x               | x               |
+| [`release-0.13`](https://github.com/prometheus-operator/kube-prometheus/tree/release-0.13) | ✗               | ✗               | x               | ✔               | ✔               | ✔               | x               | x               | x               | x               | x               |
+| [`release-0.14`](https://github.com/prometheus-operator/kube-prometheus/tree/release-0.14) | ✗               | ✗               | x               | ✔               | ✔               | ✔               | ✔               | ✔               | ✔               | x               | x               |
+| [`main`](https://github.com/prometheus-operator/kube-prometheus/tree/main)                 | ✗               | ✗               | x               | x               | ✔               | ✔               | ✔               | ✔               | ✔               | ✔               | ✔               |
```
simonpasquier, line 50: @philipgough not sure what's the correct information to put. Since we test only 1.31, 1.32 and 1.33, should we still claim that older versions are supported too?

philipgough: I think its fine. I need to think about the future of releases and testing generally anyway as the current situation isnt sustainable.

simonpasquier: yep. It probably makes sense to set a number of minor versions that we intend to support "officially" (e.g. from v1.33 to v1.29 or v1.28).
I hope that people don't run Kube 1.23 anymore :sweat_smile:

philipgough: yep, the problem was, we have not, and are not cutting releases on schedule so we were having a few versions missing from the matrix. 

I think we can have a chat about it in the office hours call

simonpasquier: definitely!

PR #2669, opened 2025-06-02
Add component label to k8s-control-resources

haskjold:
## Description

Add component label to k8s-control-resources. 

I would like to filter the resources in the generated manifests based on the `component` label and a bunch of them was missing this label. This PR together with [PR 2667](https://github.com/prometheus-operator/kube-prometheus/pull/2667) and [PR 2668](https://github.com/prometheus-operator/kube-prometheus/pull/2668) adds the missing labels.

## Type of change

## Changelog entry

```release-note
Add component label to k8s-control-resources.
```

### DIFF
--- a/jsonnet/kube-prometheus/components/k8s-control-plane.libsonnet
+++ b/jsonnet/kube-prometheus/components/k8s-control-plane.libsonnet
@@ -7,7 +7,7 @@ local defaults = {
   commonLabels:: {
     'app.kubernetes.io/name': 'kube-prometheus',
     'app.kubernetes.io/part-of': 'kube-prometheus',
-    'app.kubernetes.io/component': 'control-plane',
+    'app.kubernetes.io/component': 'kubernetes',
   },
   mixin:: {
     ruleLabels: {},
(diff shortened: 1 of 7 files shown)


jsonnet/kube-prometheus/components/k8s-control-plane.libsonnet
```diff
@@ -7,6 +7,7 @@ local defaults = {
   commonLabels:: {
     'app.kubernetes.io/name': 'kube-prometheus',
     'app.kubernetes.io/part-of': 'kube-prometheus',
+    'app.kubernetes.io/component': 'prometheus',
```
simonpasquier, line 10: I'm not sure that 'prometheus' is the right component here since we're looking at control plane services.

jsonnet/kube-prometheus/components/k8s-control-plane.libsonnet
```diff
@@ -7,6 +7,7 @@ local defaults = {
   commonLabels:: {
     'app.kubernetes.io/name': 'kube-prometheus',
     'app.kubernetes.io/part-of': 'kube-prometheus',
+    'app.kubernetes.io/component': 'control-plane',
```
simonpasquier, line 10: thinking more about it, some of the resources are about Kubelet and CoreDNS. Maybe this suggestion is more generic (noting that the `app.kubernetes.io/name` labels already tell about the specific component)?
```suggestion
    'app.kubernetes.io/component': 'kubernetes',
```

PR #2684, opened 2025-07-06
change: relax kube_.+_created to allow namespaces

rexagod:
Signed-off-by: Pranshu Srivastava <rexagod@gmail.com>

## Description

KSM v2.16 introduced look around support for regexes (https://github.com/kubernetes/kube-state-metrics/pull/2616), and since namespaces are an extremely common native resource, it might make sense to have their creation metrics OOTB for downstream consumers.

## Type of change

- [X] `CHANGE` (fix or feature that would cause existing functionality to not work as expected)

## Changelog entry

```release-note
Replaces the `^kube_.+_created$` in KSM's metrics deny-list with `^kube_(?=namespace).*_created$` to allow generating namespace creation metrics.
```
***

`generate-versions.sh` seems to return `null` for KSM, but I can update `versions.json` if needed.


rexagod:
If this still seems odd, I'll try to put together a more makeshift solution downstream and close this out.

simonpasquier:
I suspect that cluster-monitoring-operator is the only downstream user for his addon. I'm not sure that it's worth keeping it in this repository given that it's not documented and it provides no configuration option. If we want to keep it here, I'd advocate for 1) adding arguments to customize the deny list and 2) minimal documentation.

WDYT @slashpai @philipgough?

rexagod:
Ping @simonpasquier for a look here 🙂

simonpasquier:
Thanks! https://github.com/prometheus-operator/kube-prometheus/pull/2684#discussion_r2236955741 is still valid.

rexagod:
So just to make sure I'm understanding this correctly, do you mean instead of `ksmConfig` this should use `values` so it can be integrated into the [`values`-based model](https://github.com/openshift/cluster-monitoring-operator/blob/49eed1d8198cf8b3a6d6c3c6648687ea33165bcd/jsonnet/main.jsonnet#L101) used in CMO?

PR #2826, opened 2026-05-06
fix(monitoring-alerting): use max by() in KubeQuota rules to avoid many-to-many errors

sebastiangaiser:
## Description

During kube-state-metrics rolling restarts (e.g. node reboots), two pods temporarily scrape the same metrics. The upstream rules use max without(instance, job, type) which preserves pod/container/endpoint/service labels — these differ between the two pods, producing duplicate series on the join's right-hand side. Prometheus then rejects the query with "many-to-many matching not allowed".

## Type of change

## Changelog entry

```release-note
use max by() in KubeQuota rules to avoid many-to-many errors
```


sebastiangaiser:
Ah, these are coming from https://github.com/kubernetes-monitoring/kubernetes-mixin. Will open the PR there. Sorry for the noise

PR #2836, opened 2026-05-14
Improve Kubernetes alert rules to reduce alert fatigue

Nk-dir:
## Summary

This PR adds custom Prometheus alert rules to improve Kubernetes observability and reduce alert fatigue.

## Changes
- Added CrashLoopBackOff alert improvements
- Excluded system namespaces from alerts
- Added high CPU usage alert
- Improved alert annotations for readability

## Motivation
Default alert rules can generate noisy alerts from system namespaces and temporary deployment events. These changes help improve signal-to-noise ratio for operational monitoring.

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,3 +124,13 @@ If you have any questions or feedback regarding kube-prometheus, join the [kube-
 ## License
 
 Apache License 2.0, see [LICENSE](https://github.com/prometheus-operator/kube-prometheus/blob/main/LICENSE).
+
+## Custom Alert Improvements
+
+Added custom Prometheus alert rules to reduce noisy Kubernetes alerts and improve observability.
+
+### Features
+- Excludes system namespaces from crash loop alerts
+- Detects high CPU usage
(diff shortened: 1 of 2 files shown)


simonpasquier (commented):
The files under the `manifests/` directory are generated from the jsonnet code.
About the 2 alerting rules that you want to add:
* `CriticalPodCrashLooping` is similar to `KubePodCrashLooping` which comes from https://github.com/kubernetes-monitoring/kubernetes-mixin
* `HighCPUUsage` is too generic and doesn't follow the [principle](https://prometheus.io/docs/practices/the_zen/#symptom-based-alerts-for-paging-cause-based-for-troubleshooting) of alerting on symptoms and not causes.

See [examples/prometheus-additional-alert-rule-example.jsonnet](https://github.com/prometheus-operator/kube-prometheus/blob/main/examples/prometheus-additional-alert-rule-example.jsonnet) for adding new rules to the default manifest bundle.

PR #2846, opened 2026-06-17
docs: document the conflict with an existing metrics-server

### LINKED ISSUE #926: Document guidance to remove metrics-server before installing kube-prometheus
**What is missing?**

Some (?) deployments of Kubernetes come pre-installed with a metrics-server implementation in order to provide metrics API to `kubectl` (e.g. `kubectl top ...`). It may be wise to uninstall this existing metrics-server installation before installing kube-prometheus. Documenting this as an installation step would be useful.

**Why do we need it?**
Documenting a recommendation to remove metrics-server both encourages a clean cluster state and an understanding that parts of kube-prometheus will interact with / replace parts of metrics-server, and that with kube-prometheus, metrics-server is no longer required.

**Environment**

* kube-prometheus version:

    `release-0.7`

Raised following discussions in #kube-prometheus Slack.

yugstar:
## Description

Adds a troubleshooting section documenting the conflict between kube-prometheus and a pre-existing `metrics-server` — a recurring footgun in #926 (multiple users reported failed installs and, worse, control-plane metrics going down).

kube-prometheus bundles the [Prometheus Adapter](https://github.com/kubernetes-sigs/prometheus-adapter), which registers the `v1beta1.metrics.k8s.io` APIService. A cluster can only have one provider of that APIService, so:

- installing on top of an existing metrics-server fails with `Error from server (AlreadyExists): ... "v1beta1.metrics.k8s.io" already exists`, and
- uninstalling kube-prometheus can delete the shared APIService and break `kubectl top` / HPA.

The new section explains this and gives the removal steps (Helm uninstall, or the manifest cleanup commands that users in the issue verified). Placed in `docs/troubleshooting.md`; happy to also add a short note to the README Prerequisites if maintainers prefer.

Closes #926

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -47,3 +47,35 @@ By default, kubeadm will configure kube-proxy to listen on 127.0.0.1 for metrics
 1. Before cluster initialization, the config file passed to kubeadm init should have KubeProxyConfiguration manifest with the field metricsBindAddress set to 0.0.0.0:10249
 2. If the k8s cluster is already up and running, we'll have to modify the configmap kube-proxy in the namespace kube-system and set the metricsBindAddress field. After this kube-proxy daemonset would have to be restarted with
    `kubectl -n kube-system rollout restart daemonset kube-proxy`
+
+## Conflict with an existing metrics-server
+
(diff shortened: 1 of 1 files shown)


simonpasquier (approved):
thanks!

